Issue #4, opened 2014-09-22
Collect metadata on each program


briandfoy:
Collect meta data for all the programs. Store them in a consistent way in the program for now.

See the bin/factor program for a start. See the top =pod section

svatsan:
@briandfoy close this issue? Looks like this was already handled via #8. Unless you're looking for something more

briandfoy:
I'd like to fill in the holes in many of the entries.

svatsan:
Fair enough, I can perhaps take a look around sometime next week to see what's missing

jgpuckering:
Brian,

I whipped up a script to extract the metadata and print it as a tab-separated table for review in Excel.  Here's what I found:

```
Missing Authors:  tar, tty

Missing email addresses:
  Brad Appleton       test
  Jeffrey S. Haemer   addbib, asa, clear, colrm, mkfifo, tsort, uname
  Joshua Gross        date
  Louis Krupp         mkfifo
  Michael Mikonos     base64, hexdump, morse
  Paul Grassie        grep
  dkulp               ar, file, time

Non-standard email:
  Nathan Scott Thompson, "quimby at city-net dot com"
  (in deroff, strings)

Missing license:
  arch, dc, expr, hexdump, id, kill, mimedecode, 
  ppt, strings, sum, tar, tty, whois, xargs
```

Some of these holes can be plugged.  For example, I found an email for Brad Appleton in CPAN -- but without verification it might be pointless to add it.  But I'd be happy to go through that process.

Nathan Scott Thompson has a email but it's inconsistent with the rest.  Easy to fix.

Speaking of inconsistencies, the Description for rot13 starts with a propercase word.  All others are lowercase.  I'd like to change it to lowercase.

I didn't actually validate the License info.  I suspect you already did that, but if you'd like an independent review checking the metadata entry against the POD in the file, I can do that.

As for the missing licenses, most show as perl_5 in MetaCPAN -- probably the default.  Not sure whether the files should be updated accordingly.  Most of these are so old that it seems likely they never had any license information.  Is it valid to add it now???  At Cognos and then IBM I used to do compliance work, including some related to copyright.  It's a complex area.  Nonetheless, it seems to me that if the code is published via MetaCPAN and a license is listed on the page, and most scripts include said license, then all should.  What are your thoughts on that?

As for scripts that are missing metadata, I only found two:  base64 and hexdump.  I have an update for those.

[...]

PR #8, opened 2014-10-04
Add metadata to each program

timgimyee:
Issue #4.

### DIFF
--- a/bin/addbib
+++ b/bin/addbib
@@ -1,5 +1,14 @@
 #!/usr/bin/perl
 
+=for info
+
+Name: addbib
+Description: create or extend a bibliographic database
+Author: Jeffrey S. Haemer
+License: 
+
+=cut
+
 use strict;
 use Getopt::Std;
 use vars qw($opt_a $opt_p);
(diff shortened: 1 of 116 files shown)


kentfredric:
It has come to my attention that this POD in this pull request is formatted incorrectly.

According to POD spec, 


>        That is, with "=for", you can have only one paragraph's worth of text
>        (i.e., the text in "=foo targetname text..."), but with "=begin
>        targetname" ... "=end targetname", you can have any amount of stuff in
>        between. (Note that there still must be a blank line after the
>        "=begin" command and a blank line before the "=end" command.)

So that means the following:

```perl
=for foo

bar

=cut

```

Is parsed as:

```perl
{ type => "for", formatname => "foo", text => "" }
{ type => "text", text => "bar" }
{ type => "cut" }
```

Where what you meant  to do was

```perl
=begin foo

bar

=end foo

=cut
```

Which communicates:

```perl
{ type => "begin", formatname => "foo", text=>"bar\n" }
{ type => "cut" }
```

Subsequently, this mistake renders the "metadata" as actual documentation: 

https://metacpan.org/pod/touch


Because a "bank event" ( ie: a line without content ) marks the end of the "Paragraph" which is the "=for" directive.

Issue #12, opened 2014-12-20
find: does find -mtime work on Windows?


briandfoy:
http://www.nntp.perl.org/group/perl.ppt/2006/02/msg38.html

An old message discusses a problem where hours seemed like days.

kal247:
'find -mtime' does work on Windows.

briandfoy:
@kal247 does that mean I can close this issue?

PR #57, opened 2018-10-12
Add -amin, -cmin and -mmin to `find` command

timgimyee:
For #12.  Modify output of find2perl to support -amin, -cmin and -mmin.

### DIFF
--- a/bin/find
+++ b/bin/find
@@ -18,52 +18,109 @@ License: perl
 
 use strict;
 
-my(
-   $temp_file,  # file to capture the output of find2perl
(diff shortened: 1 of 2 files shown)


briandfoy:
This looks good I think, but I want to consider that `system` call a bit more before I merge it.

briandfoy:
One thing I'd like to add is a search for find2perl. If it's not in the PATH, give the user a message that they need to install App::find2perl. Even though Perl Power Tools has that as a prerequisite someone might copy the script somewhere.

PR #60, opened 2018-10-23
Add sort tests

shlomif:
Hi! I changed the licence to Artistic 2.0 now. Note that AFAIK (and IANAL) the https://tldrlegal.com/license/mit-license can be sublicensed into the https://tldrlegal.com/license/artistic-license-2.0-(artistic) even without asking me. Anyway, I also refactored the code a little. Please look into merging these.


briandfoy:
I'd prefer that you'd start over with that licensed file so the unacceptable license doesn't show up in a commit. Everything else looks good.

PR #64, opened 2018-12-21
Fix a format string bug in "cat -n".

shlomif:
See https://en.wikipedia.org/wiki/Uncontrolled_format_string . The
output was wrong if given %d / etc. as well.


shlomif:
Hi! Any news?

shlomif:
Hi! Do you need anything or can this be merged?

shlomif:
Thanks!

PR #65, opened 2019-07-25
Sort --head and --tail flags

shlomif:
These use heaps/PQs to speed up the sort | head/tail idioms. A nonstandard feature that I prototyped in PPT.

### DIFF
--- a/bin/sort
+++ b/bin/sort
@@ -69,7 +69,10 @@ Getopt::Long::config('bundling');  # -cmu == -c -m -u
 
 package MyElem;
 
-use parent 'Heap::Elem';
+use vars qw/ @ISA /;
+BEGIN {
+@ISA = ('Heap::Elem');
+};
 
 sub new {
     my $self = shift;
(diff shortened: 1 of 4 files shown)


bin/sort
```diff
@@ -91,7 +120,26 @@ sub _sort_file {
...
     # see big ol' mess below
     _make_sort_sub($opts);
-
+    {
+        local $^W;
+        *MyElem::_s = ($MyElem::opt_K = $opts->{K}) ? \&sort1 : \&sortsub;
+    }
+    my $pq;
+    if (defined $head or defined $tail) {
+        require Heap::Fibonacci;
```
briandfoy: With PerlPowerTools, we have to be mostly self contained. If it's not in Perl core, we need to stay away from it.

shlomif: I see. Well, these modules won't get used as long as the flags are not specified. This branch was also just a proof-of-concept for a similar feature in gnu coreutils or whatever. Feel free to close this pull req - https://www.shlomifish.org/philosophy/philosophy/putting-cards-on-the-table-2019-2020/putting-cards-on-the-table-2019-2020.raw.html#do_and_let_do__live_and_let_live and [Hakuna Matata](https://duckduckgo.com/?q=hakuna+matata&atb=v140-1&ia=videos)

PR #66, opened 2019-08-17
Fix multiple filenames in ARGV in tac with a test.

shlomif:
There was a mix of sysseek and tell.

### DIFF
--- a/t/tac.t
+++ b/t/tac.t
@@ -3,7 +3,7 @@
 use strict;
 use warnings;
 
-use Test::More tests => 2;
+use Test::More tests => 3;
 
 sub _lines2re
 {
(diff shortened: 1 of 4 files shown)


t/tac.t
```diff
@@ -0,0 +1,187 @@
...
+        flags => [qw/ -rs "[az]" /],
+        lines => [ split /\n/, <<'EOF'],
+threeatwoaonea
+EOF
+    }
+);
+
+__END__
+
+=head1 COPYRIGHT & LICENSE
```
briandfoy: You need to license everything in the same way that the entire package is licenses. In this case, it's exactly the Artistic License 2.0 (as seen in LICENSE). It can't be a future license that does not exist.

shlomif: hi @briandfoy ! I hope you don't take it the wrong way but you seem a little too pedantic and stubborn and bikeshedding here - see http://shlomifishswiki.branchable.com/Never_Try_to_Please_Everyone/ . While [IANAL](https://en.wikipedia.org/wiki/IANAL) but it is not inconceivable that one day there will be The Artistic License 2.1 or 3.0 or whatever. You can be a bit more tolerant. [Hakuna matata](https://duckduckgo.com/?q=hakuna+matata&atb=v140-1&ia=about) .

briandfoy:
Okay. I’ll close this issue. Since you decided to make personal attacks, I won’t accept any further contributions from you.

shlomif:
@briandfoy : hi! First of all: I'm sorry if I mispoke. Anyway, I didn't intend to imply there was anything wrong with your character, just that I disagreed with your criticism of my licensing choice, which I hoped you could avoid in the future (because people can change). I can license my changes as Artistic License version 2.0 (only) from now on in order to avoid future disagreements.

( Note that I requested a friend to try to comment here , and he agreed but he didn't get to doing that in time, which caused a delay in writing this comment. I'm sorry for that as well. )

PR #67, opened 2019-08-21
Fixes for bin/robots

jwrightecs:
The block checking for the Curses module was always calling exit, so bin/robots never ran;  correcting that, bin/robots was not importing Curses constants, instead using package globals created with use vars which were unset.


briandfoy:
Thanks!

PR #77, opened 2019-10-20
update asa

oWnOIzRi:
enable warnings and use any instead of grep in boolean context

### DIFF
--- a/bin/asa
+++ b/bin/asa
@@ -13,10 +13,12 @@ License:
 
 
 use strict;
+use warnings;
+use List::Util qw/ any /;
 
-exit 1 if grep {!-r} @ARGV;		# traditional
+exit 1 if any {!-r} @ARGV;		# traditional
 
-if (grep /-/, @ARGV) {
+if (any {/-/} @ARGV) {
 	$0 =~ s(.*/)();
 	warn "usage: $0 [filename ...]\n";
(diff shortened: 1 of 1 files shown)


bin/asa
```diff
@@ -13,10 +13,12 @@ License:
 
 
 use strict;
+use warnings;
+use List::Util qw/ any /;
```
briandfoy: The question is if loading a module saves a significant amount of time over using a slightly inefficient grep. Can you benchmark both ways and see which one is faster?

oWnOIzRi: Loading the module caused the 5.8 travis build to fail which I note is a requirement so I reverted back to the grep block form in the second commit.  So ultimately the only thing that changed was I turned on warnings and changed the grep from expression to block form.

briandfoy: That's odd because List::Util was core in 5.7.3.

For the warnings, I have the same comment as in #78. I'm not sure about shipping developer settings to end users.

oWnOIzRi: 'any' is only listed as a suggested addition in 5.8 docs.

Let me know if you make a decision, I'm happy to help in any way I can.

briandfoy:
So far we haven't added warnings to the scripts, so I'm deferring this.

PR #78, opened 2019-10-20
update arithmetic

oWnOIzRi:
enable warnings in arithmetic

### DIFF
--- a/bin/arithmetic
+++ b/bin/arithmetic
@@ -13,6 +13,7 @@ License: perl
 
 
 use strict;
+use warnings;
 use integer;
 
 my ($VERSION) = '1.2';


bin/arithmetic
```diff
@@ -13,6 +13,7 @@ License: perl
 
 
 use strict;
+use warnings;
```
briandfoy: Let me think about this for a bit. These are programs designed for end users who might not even know that Perl is the language behind them. As such, warnings aren't a good interface for them. Warnings are a good developers tool and we should ensure that the programs are warnings safe, but shipping end products with developer settings is a different beast.

briandfoy:

bin/arithmetic
briandfoy: I've decided for the time being that we won't enable warnings. The main problem is that our audience isn't necessarily Perl programmers or even developers. Something be warnings safe now might not be with a future version. The intended user probably won't care or understand the warning.

We can test and develop with warnings on, but not ship with warnings.

Issue #73, opened 2019-10-20
Get rid of indirect object notations


briandfoy:
Here's a bunch of violations:

https://kritika.io/users/briandfoy/repos/7253941177493998/snapshots/1/violations?rule=Objects::ProhibitIndirectSyntax

It's code that puts the method first (as C++ would):

     $t0 = new Benchmark;

Turn this into the preferred syntax:

    $t0 = Benchmark->new;

That's it.

The affected files. Do as many as you like:

- [x] bin/file
- [x] bin/robots
- [x] bin/look
- [x] bin/ar
- [x] bin/whois
- [x] bin/diff
- [x] bin/time
- [x] bin/mail
- [x] bin/pr
- [x] bin/rev
- [x] bin/tac

https://kritika.io/users/briandfoy/repos/7253941177493998/snapshots/1/files/bin/file?expand=104#L104

IvanLudvig:
I'm going to try to solve this issue

PR #79, opened 2019-10-21
Fixed indirect object notation

IvanLudvig:
Fixed 11 files so the program uses the preferred syntax instead of indirect object notation. Issue #73

### DIFF
--- a/bin/ar
+++ b/bin/ar
@@ -225,7 +225,7 @@ sub extractMember {
 				   oct($attr->[4]), int($attr->[1]));
     }
 
-    my $out = new FileHandle ">$name" or die "$name: $!\n";
+    my $out = FileHandle->new ">$name" or die "$name: $!\n";
     binmode($out);
     $out->print($attr->[6]);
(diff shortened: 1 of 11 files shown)


bin/ar
```diff
@@ -225,7 +225,7 @@ sub extractMember {
 				   oct($attr->[4]), int($attr->[1]));
     }
 
-    my $out = new FileHandle ">$name" or die "$name: $!\n";
+    my $out = FileHandle->new ">$name" or die "$name: $!\n";
```
briandfoy: Let's put some parens around the argument list (for this line and all the others like it)

FileHandle->new( ">$name" )

IvanLudvig: Added the parens

briandfoy:

briandfoy:
I've merged this, but manually because I had to rebase onto previous pull requests. See 1302f660e0613e3733d3ec41fc90ec7c66fb08ee.

Thanks!

Issue #74, opened 2019-10-20
Use lexical loop iterators


briandfoy:
There are a bunch of `for` and `foreach` loops that don't use lexical loop variables:

https://kritika.io/users/briandfoy/repos/7253941177493998/snapshots/1/violations?limit=20&page=1&rule=Variables%3A%3ARequireLexicalLoopIterators

For example:

    foreach $x ( ... )

Instead, that should be:

    foreach my $x ( ... )

Many of these program are not yet strict safe, but this is a good step to getting closer to that.

Do as many or as few of the problem files as you like. For Hacktoberfest, it doesn't matter how big or complete the pull request is. :)

mohawk2:
This seems to be fixed by the attached PR? Can it be closed?

PR #80, opened 2019-10-22
fix lexical loop iterators

interlab:
Fixed Issue #74


briandfoy:

briandfoy:
This is merged but I did it manually because I fixed a few formatting things. These weren't your problem but that I noticed in a few files that you changed.

Thanks!

PR #83, opened 2020-02-19
Check-in initial version of rot13

marked:
Created rot13 filter script
Checked version installed from bsd-games, and it ignores any arguments.


briandfoy:
Each of the programs has some metadata at the top to identify the author and whatnot. Can you add that?

briandfoy:
I'll fix up the meta data.

PR #84, opened 2020-05-22
Fix issue with `NULL` filehandle in patch.

hakonhagland:
Refer to [Saving a reference to a localized filehandle. How does it work?](https://stackoverflow.com/q/61952171/2173773)

Currently patch saves an reference to a localized NULL filehandle in the object hash in its `bless()` method. But when the `bless()` method returns the original value of `*NULL` is restored (which is empty). The value is restored since `*NULL` was localized. This means that when `patch` later calls `apply()` and tries to print to the  `NULL` handle, we get a warning about trying to print to a closed filehandle.

The correct behavior should be to store a copy of `*NULL` (not a reference to `*NULL`)  or to not localize `*NULL` in the first place..

### DIFF
--- a/bin/patch
+++ b/bin/patch
@@ -434,8 +434,8 @@ sub bless {
 
     # Open output file.
     if ($self->{check}) {
-        $self->{o_fh} = \*NULL;     # output filehandle
-        $self->{d_fh} = \*NULL;     # ifdef filehandle
+        $self->{o_fh} = *NULL;     # output filehandle
+        $self->{d_fh} = *NULL;     # ifdef filehandle
     } else {
         local *OUT;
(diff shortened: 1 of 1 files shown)


mohawk2:
This may be none of my business, sorry if so! But I think the AppVeyor CI is now working on `master`, so rebasing this change (which seems sound otherwise) would be valuable?

(On my projects I usually first aim to "capture the error" by making a test that fails, then write the code to fix it - how hard would it be to add a test here that demonstrates this error?)

mohawk2:
@hakonhagland Thanks for the thumbs up! So... rebase this and add a test? :-)

briandfoy:
I've rebased this in my clone and created a new PR in #91. It's the same commit though.

Issue #85, opened 2020-07-05
Modulinos?


mohawk2:
It seems to me all the PPTs should be modulinos?

briandfoy:
That would complicate things quite a bit. These should be single file things that people can grab individually.

mohawk2:
It seems I don't understand modulinos, then. I thought that described Perl code that could be grabbed individually and just run as a command-line script, but if loaded into another program would make its "main" function instead be callable? The changes required to implement this would mostly involve inserting `package PerlPowerTools::cat_or_whatever; sub main {` and `}` and a `main(@ARGV) unless caller`.

If that understanding were correct, then the various scripts in this distro could also be utilised by e.g. pure-perl-make to avoid calling the shell at all for supported tasks, which would make things much faster, particularly on Windows. Spoiler: I am in fact intending to make pure-perl-make do that. I'd hoped these scripts would make that easy.

Can you help me understand why this would be complicated?

briandfoy:
Ah, I see what you are saying. I guess that could work if someone wanted to do the work. But, the tests still need to check them as scripts.

If you'd like to do that, pick one program and let's try it out.

mohawk2:
Of course you're right that starting with one is a good idea.

My approximate plan will be to start also from using https://metacpan.org/pod/Modulino::Demo, while also probably making a small test framework that makes testing the scripts/modules much easier: I'm thinking of a data-driven approach allowing one to just express fixtures and test-cases, that would run the tests against both the script incarnation and the module incarnation with no extra effort.

briandfoy:
I'm not closing this as a rejection. I've answered the question and if you want to move forward, send a PR. Just don't go crazy with frameworks and such though :)

PR #86, opened 2020-07-05
Fix tests for Windows, also "perl in space"

mohawk2:
Just as a by-the-by, a lot of this wouldn't be necessary if the scripts were all modulinos (#85) as testing would be much easier.

### DIFF
--- a/t/find.t
+++ b/t/find.t
@@ -32,8 +32,8 @@ HERE
 diag( "find2perl is at <$find2perl>" );
 
 
-subtest check_find2perl => sub { check_find2perl( $find2perl ) };
-subtest check_find      => \&check_find;
+subtest check_find2perl => \&check_find, "find2perl", $find2perl;
+subtest check_find      => \&check_find, "find", $find;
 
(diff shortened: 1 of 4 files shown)


t/find.t
```diff
@@ -104,17 +104,17 @@ sub check_find2perl {
 	my( $find2perl ) = @_;
 
 	ok( -e $find2perl, "find2perl exists at $find2perl" );
-	ok( -x $find2perl, "find2perl is executable $find2perl" );
+	ok( -x $find2perl, "find2perl is executable $find2perl" ) if $^O ne 'MSWin32';
```
briandfoy, line 107: Can you use SKIP blocks where instead?

briandfoy:
Even if these modules were modulinos, I'd still want to do the interface tests just as you see here.

t/find.t
mohawk2: Done. I also unified the nearly-identical `check_find{,2perl`.

briandfoy:
I think this looks generally good, but give me a chance to test it on my side.

PR #88, opened 2020-08-11
open2 frequently hangs on Windows, so use IO redirect instead

chorny:


mohawk2:
This does not seem like a good way of solving this problem. You could at least use `File::Temp`? Is this a sign of a bug in open2?

briandfoy:
Yeah, I'll have to come up with another solution.

Issue #89, opened 2020-08-20
PerlPowerTools fails tests on Windows 10 or Server


iamlgh:
When I try to install PerlPowerTools on Strawberry Perl (v5.20.2), the t\find.t test fails.

I traced it to two issues:
1) -x file test is invalid (and never returns true) on Windows.
2) The file slashes don't match when comparing results on Windows.

Here is the excerpt from the build log:
```
t/find.t ..... # find2perl is at <C:\Strawberry\perl\site\bin\find2perl>

    #   Failed test 'find2perl is executable C:\Strawberry\perl\site\bin\find2perl'
    #   at t/find.t line 107.
    # Looks like you failed 1 test of 3.
t/find.t ..... 1/?
#   Failed test 'check_find2perl'
#   at t/find.t line 35.

    #   Failed test 'find is executable blib\script\find'
    #   at t/find.t line 115.
    # Looks like you failed 1 test of 3.

#   Failed test 'check_find'
#   at t/find.t line 36.
# Files are
#  C:\Users\sdklsh\AppData\Local\Temp\perlpowertools-find-R6p1wK0i\a\b\c\20.txt
#  C:\Users\sdklsh\AppData\Local\Temp\perlpowertools-find-R6p1wK0i\d\40.txt
#  C:\Users\sdklsh\AppData\Local\Temp\perlpowertools-find-R6p1wK0i\e\f\60.txt
#  C:\Users\sdklsh\AppData\Local\Temp\perlpowertools-find-R6p1wK0i\g\h\i\80.txt
    # C:\Users\sdklsh\AppData\Local\Temp\perlpowertools-find-R6p1wK0i\a\b\c\20.txt  a:1597904345  m:1597904345
    # C:\Users\sdklsh\AppData\Local\Temp\perlpowertools-find-R6p1wK0i\d\40.txt  a:1597903145  m:1597903145
    # C:\Users\sdklsh\AppData\Local\Temp\perlpowertools-find-R6p1wK0i\e\f\60.txt  a:1597901945  m:1597901945
    # C:\Users\sdklsh\AppData\Local\Temp\perlpowertools-find-R6p1wK0i\g\h\i\80.txt  a:1597900745  m:1597900745

    #   Failed test 'Found files with `C:\Strawberry\perl\bin\perl.exe blib\script\find C:\Users\sdklsh\AppData\Local\Temp\perlpowertools-find-R6p1wK0i -type f`'
    #   at t/find.t line 72.
[...]

PR #102, opened 2021-09-07
Add whoami command

ersen0:


bin/whoami
```diff
@@ -0,0 +1,50 @@
...
+Author: Oğuz Ersen, oguzersen@protonmail.com
+License: perl
+
+=end metadata
+
+=cut
+
+use strict;
+
+my $user = getpwuid($>);
```
briandfoy, line 18: Note that getpwuid doesn't work on Windows, so you'll need another way to handle that for that platform.

bin/whoami
```diff
@@ -0,0 +1,50 @@
...
+=head1 SEE ALSO
+
+whoami(1), id(1)
+
+=head1 COPYRIGHT and LICENSE
+
+Copyright (c) by Oğuz Ersen 2021.
+
+This program is free and open software. You may use, modify, distribute,
+and sell this program (and any modified variants) in any way you wish,
```
briandfoy, line 47: Note that the declared license in the metadata is "perl", so there are some restrictions there.

briandfoy (changes requested):

bin/whoami
briandfoy: Or, maybe it's more complicated:

https://stackoverflow.com/questions/3526420/how-do-i-get-the-current-user-in-perl-in-a-portable-way

ersen0: I am considering:

    my $user = $^O eq 'MSWin32' ? $ENV{USERNAME} : getpwuid($>);

I am not sure if it will work since I don't have access to a Windows machine right now.

bin/whoami
ersen0: I have copied and pasted this from other commands, e.g. `bin/awk` has the same. Is it OK if I change this section with (also copied and pasted from other commands, i.e. `bin/fold`):

    This code is freely modifiable and freely redistributable under Perl's
    Artistic License.

bin/whoami
briandfoy: Don't settle on anything until you know it will work on Windows. Note the strategies that I noted and what's mentioned in that Stackoverflow post.

bin/whoami
briandfoy: It would be better to simply say "This code is licensed under the Artistic License 2."

bin/whoami
ersen0: How about:

```
my $user;

if ($^O eq 'MSWin32') {
    $user = getlogin() || $ENV{USERNAME};
}
else {
    $user = getpwuid($>);
}

print "$user\n";
```

Or should I use `Win32::LoginName`?:

```
my $user;

if ($^O eq 'MSWin32') {
    require Win32;
    $user = Win32::LoginName;
}
else {
    $user = getpwuid($>);
}

print "$user\n";
```

briandfoy: I don't know the answers to these. You'll need to try these out in various situations on Windows to see if they work.

But, remember that one of the goals is that we only use core modules. Optionally using a non-core module is fine if the operation doesn't depend on that.

bin/whoami
```diff
@@ -0,0 +1,57 @@
...
+
+=cut
+
+use strict;
+
+my $user;
+
+if ($^O eq 'MSWin32') {
+    require Win32;
+    $user = Win32::LoginName();
```
briandfoy, line 22: handle the case where Win32::* is not available. See the `getlogin()` that I showed you earlier. Remember, you are mostly limited to core modules.

briandfoy (changes requested):

bin/whoami
ersen0: I changed this part with `$user = getlogin() || $ENV{USERNAME};`, Is this enough or are there edge cases where both `getlogin()` returns empty and `USERNAME` environment variable is not available? I can test it as soon as I have access to a Windows machine but it may take a while.

briandfoy: Once you get a chance to work with this on Windows, see what works.

But, if you don't have access to Windows, is there some other reason you'd like this program to be in PerlPowerTools? Is there another non-unix system you are targeting that doesn't already have a `whoami`?

ersen0: I just saw it was missing and wanted to contribute (: This can be closed without merging if it is not suitable for PerlPowerTools. Thanks for your guidance and patience.

briandfoy:
No, it's a totally appropriate addition. We just have to figure out how to get it right on Windows. Major pain in the butt (see #89 for example).

briandfoy:
I've merged this manually and did a little work on it. It will be part of the next release. Thanks,

PR #104, opened 2022-02-01
packer.pl v1.21 + perldoc

kal247:
- better doc in packer.pl header
- added perldoc (with custom search path) as a tool


briandfoy:
This looks good at first glance. That's a pretty clever solution to the perldoc problem. I'm merging this, and maybe the next step is to use GitHub actions to create the releases for different platforms. If you want to play with that, you might look at https://github.com/briandfoy/github_actions/blob/master/perl-module-release.yml to see what I've been playing with. Instead of the `make dist` step, you'd do the `pp` stuff and upload that (now #105)

PR #129, opened 2023-03-07
od: exit(1) for bad options. garbage collect -a and -h

mknos:
* match GNU od: exit with error code and do not continue if bad options are detected
* no need for -h (help) flag; usage already printed
* remove unsupported -a flag mentioned in usage string and POD


briandfoy (approved):

briandfoy:
This patch is applied, but I had already applied part of it in a previous commit so the actual diff may look weird. No big whoop.

thanks,

PR #130, opened 2023-03-07
hexdump tool proposal

mknos:
* reference for default output: hd command on linux
* character-dump option -c, based on linux hexdump
* reverse hexdump option -r, based on xxd command

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -45,6 +45,7 @@ bin/glob
 bin/grep
 bin/hangman
 bin/head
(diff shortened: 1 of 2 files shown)


briandfoy (changes requested):
This looks good, but we need to also add the hexdump program to MANIFEST, etc

briandfoy:
I've updated the GitHub Pages pages for this addition because there's no reason anyone would know about it.

PR #132, opened 2023-03-08
dc: remove ! command for shell

mknos:
As done in OpenBSD version of dc. This commit will print identical warning message as OpenBSD dc.
Sync TODO file.


briandfoy:
Thanks,

PR #133, opened 2023-03-09
bc: remove support for evaluating perl expressions

mknos:
The bc language has the operator '&&' but not '&'. This version of bc has a grammar extension &"STRING" which calls eval().

$ perl bc 
":)"
:)
& "syscall 23,45"
Segmentation fault (core dumped)

bc can be run under "perl -d" for debugging so the extension seems undesirable.

### DIFF
--- a/bin/bc
+++ b/bin/bc
@@ -1150,14 +1150,6 @@ if ($yyn == 80) {
 { $yyval = $yyvs[$yyvsp-0];
 last switch;
 } }
-if ($yyn == 81) {
-#line 442 "bc.y"
-{
-		  push_instr('&', $yyvs[$yyvsp-0]);
-		  $yyval = 1;
-
-last switch;
-} }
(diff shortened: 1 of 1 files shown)


briandfoy (changes requested):
I think we should also remove rule 81 from `@yyrule`. I'm tempted to then adjust all the other rules downward by 1, but it's probably better to just make that entry `undef` and leave the numbers as they are since there is some `$yyn` math I don't understand.

mknos:
Yes, we can possibly make the entry in yyrule undef. The annoying thing is, yyrule and yyname appear to be for debug/tracing only. The parser tables (including yytable) still have an entry for the grammar rule. It would be ideal to regenerate the tables without this rule but that's more complicated. The patch works the same with only the 2nd chunk (i.e. if push_instr() still happens), but my preference was to delete the extra code.

PR #134, opened 2023-03-09
od: rename option -s to -j

mknos:
* Existing -s option skips bytes at beginning of input
* NetBSD, OpenBSD and GNU versions of od(1) call this option -j because -s indicates something else according to the standard
* We can rename the option for better compatibility with shell scripts (-j is currently free)
* While here, avoid seek() if -j isn't set
* Remove unused variables $retval and $skip

Also see:
https://pubs.opengroup.org/onlinepubs/009696899/utilities/od.html

### DIFF
--- a/bin/od
+++ b/bin/od
@@ -236,7 +236,7 @@ od - dump files in octal and other formats
 
 =head1 SYNOPSIS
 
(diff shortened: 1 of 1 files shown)


briandfoy (changes requested):
Update the SYNOPSIS is the pod and I think this is good

PR #135, opened 2023-03-12
Improved version of morse game

mknos:
* Align better with bsd version: input is taken from ARGV or stdin but not read from file(s)
* Support decode and encode (previous version could not encode morse)
* Support long-form and short-form morse code, but not mixed forms (no fuzzy matching).
* Correct mapping entry for -...- to '=' (found during testing)
* Add a couple of missing mappings supported by OpenBSD version of morse
* Update POD

### DIFF
--- a/bin/morse
+++ b/bin/morse
@@ -11,48 +11,13 @@ License: perl
 
 =cut
 
-
 use strict;
-use File::Basename ();
 
-my ($VERSION) = '1.2';
+use Getopt::Std qw(getopts);
 
-my $warnings = 0;
+my ($VERSION) = '1.3';
 
-END {
-    close STDOUT || die "can't close stdout: $!\n";
-    $? = 1 if $? == 255;  # from die
-}
-
-# Print a usuage message on a unknown option.
-# Requires my patch to Getopt::Std of 25 Feb 1999.
-$SIG {__WARN__} = sub {
-
-    if (substr ($_ [0], 0, 14) eq "Unknown option") {die "Usage"};
-    $0 = File::Basename::basename ($0);
-    $warnings = 1;
-    warn "$0: @_";
-
-};
-
-$SIG {__DIE__} = sub {
-
-    $0 = File::Basename::basename ($0);
-    if (substr ($_ [0], 0,  5) eq "Usage") {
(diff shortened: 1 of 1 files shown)


briandfoy (changes requested):
With everything that you've added, you should get an author credit in the metadata and the POD

bin/morse
```diff
@@ -63,76 +28,148 @@ my %translations = reverse (
...
-        my $line;
-        s/^\s+//;                       # Leading space.
-        s/dit ?/./g; s/daw ?/-/g;       # Long to short.
-        while (/,\n/ && defined ($line = <FILE>)) {
-            $line =~ s/^\s+//;          # Leading space.
-            $line =~ s/dit ?/./g;       # Long to short.
-            $line =~ s/daw ?/-/g;       # Long to short.
-            s/,\n/ $line/;              # Combine lines.
+my %opt;
+getopts('ds', \%opt) or die("usage: $0 [-ds] [string ...]\n");
```
briandfoy, line 45: The old morse uses -d for the device. Maybe we can use something else.

bin/morse
```diff
@@ -63,76 +28,148 @@ my %translations = reverse (
...
 
 =head1 NAME
 
-morse - read morse and translate it to text
+morse -  translate text to morse code
 
 =head1 SYNOPSIS
 
-morse [files ...]
+morse [-ds] [string ...]
```
briandfoy, line 139: Let's flesh out the synopsis with actual examples

briandfoy, line 139: What about the case of reading from files, as in the original? The bad version doesn't do files, but this is a big change from the PPT version.

briandfoy (changes requested):
A few things to fix up in the docs

bin/morse
mknos: Possibly we can support both by adding a "files" flag (-f ?), where ARGV is treated as a file list instead of a string list. Then the original PPT behaviour is not lost.

briandfoy: that seems fine

briandfoy:
I've had a chance to play with this and it's all good. Thanks,

PR #136, opened 2023-03-13
xargs: report system() failure

mknos:
When GNU xargs is given an invalid command to run it prints the command and the error.

$ ls -1 a* | xargs 'cat;ls' 
xargs: cat;ls: No such file or directory

With the above input, this version of xargs fails silently (but correctly). Internally it calls system(LIST) rather than system(STRING), so the intent of the code is to bypass the shell. I think making the error string visible would avoid confusion.


briandfoy (approved):

briandfoy:
Thanks

PR #139, opened 2023-03-20
ed: implement h command

mknos:
This version of edWarn() is very unhelpful because it discards the error message and there is no way of recovering it. Standard ed implements the "h" command for displaying the saved error. Now add "h" command here too. Also in this diff:
* add a basic POD manual
* don't attempt to print a filename if ed is run with no filename argument
* remove useless 2nd argument from edWarn()
* unused variable $NumLines in edWrite()
* print errors to stdout to match behaviour of GNU ed

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -11,7 +11,6 @@ License: gpl
 
 =cut
 
-
 # What: A perl version of Unix ed editor.
 #
 #    Based on the v7 documentation using GNU ed version 0.2 as a reference.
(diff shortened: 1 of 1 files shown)


briandfoy:
I see this and briefly scanned it. I think it looks good, but give me a chance to take a second, more careful look. 

I'd like to start adding tests for these commands. This isn't holding up this PR, but when you submit these, I go through the commands manually with the current version and then with your version. I don't mind that so much because I learn a lot, but waiting on me delays the process a little.

briandfoy:
Now that we have this, perhaps we should have the H command too

PR #145, opened 2023-03-28
od: support reading from stdin

mknos:
* This version of od reads only one file, but silently ignores extra file arguments
* Read from stdin if no file argument is set, or if file is '-'
* Support -j on stdin & unseekable files by checking if seek() fails (GNU od supports -j on stdin)
* Print error string if read() fails, e.g. when reading from a directory

### DIFF
--- a/bin/od
+++ b/bin/od
@@ -17,9 +17,8 @@ use Getopt::Std;
 use vars qw/ $opt_A $opt_b $opt_c $opt_d $opt_f $opt_i $opt_j $opt_l $opt_o
 $opt_v $opt_x /;
 
-my ($offset1, $radix, $data, @arr, $len);
+my ($offset1, $radix, $data, @arr, $len, $fh);
 my ($lastline, $upformat, $pffmt, $strfmt, $ml);
-local *FH;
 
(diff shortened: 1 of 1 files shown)


bin/od
```diff
@@ -63,15 +62,26 @@ getopts('A:bcdfij:lovx') or help();
...
+else {
+    $fh = *STDIN;
+}
 
-binmode(FH);
-seek(FH, $offset1, 0) if $offset1;
+binmode $fh;
+if ($offset1 && !seek($fh, $offset1, 0)) {
+    foreach (1 .. $offset1) {
+	$len = read $fh, $data, 1;
```
briandfoy, line 75: Is there a reason to read here instead of just immediately seek to the offset?

briandfoy (commented):
One small question about the offset.

Everything else looks good.

bin/od
mknos: * The default behaviour is to seek() if we get a value for -j (L72)
* For a pipe or stdin this will first fail, and the read loop only runs for a failed seek()
* There's no special logic upfront to determine if the file supports seeking

PR #151, opened 2023-04-06
Add a readlink command

mknos:
* This can be implemented without relying on things outside of the core perl distro
* Lightly tested on my ubuntu system
* readlink() and realpath() might produce strange results on non-unix systems but support code could be added in here later if needed
* Hopefully the POD text makes sense

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -78,6 +78,7 @@ bin/printf
 bin/pwd
 bin/rain
 bin/random
+bin/readlink
 bin/rev
 bin/rm
 bin/rmdir
--- a/bin/readlink
+++ b/bin/readlink
@@ -1,0 +1,109 @@
+#!/usr/bin/perl
+
+use Cwd qw(realpath);
+use Getopt::Std qw(getopts);
+
+my $VERSION = '1.0';
+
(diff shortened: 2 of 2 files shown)


briandfoy:
This would be great, but since Perl Power Tools is about providing tools to non-Unix systems, I'd like to figure out where this will break on Windows and include some advice there. In general, Windows is the most important target for the project. I don't know enough about Windows to have any guidance here, but I'm going to ask around to get some feedback.

mknos:
In 2003 I might have run a WinXP laptop with perl running under cygwin. Cygwin itself seems to [barely] deal with symlinks, but I don't remember using the readlink() function under "native" win32 perl, e.g. activeperl.
I'm not sure if the existing commands chgrp, chown and ln would do anything in activeperl.

briandfoy:
For this to go in, I think we have to answer yes to "is there any system that has symlinks but no readlink". If there's such a system, no problem.  Otherwise I don't think anyone needs it.

briandfoy:
I'm going to close this for now, but if we come up with answers for the questions I raised, we can re-examine this.

PR #152, opened 2023-04-11
bc: allow Math::BigFloat again

mknos:
* This bc implementation previously supported Math::BigFloat but it was commented out
* Add a command option (-b) to interpret number literals as BigFloat (based on commented code)
* Older perls (e.g. 5.8.8) may not support BigFloat so disable it by default
* Remove outdated comment that Math::BigFloat doesn't support hex literals
* Standard bc is not required to support hex literals, and we still don't
* Remove outdated comment that Math::BigFloat doesn't support ** operator

Output:
$ perl bc   # default 
x = 12341234123489712364.123  
1.23412341234897e+19
x ** 2
1.52306059690787e+38
quit
$ perl bc -b   # new flag
x = 12341234123489712364.123
12341234123489712364.123
x ** 2
152306059690786889006379355337832937737.559129
quit

### DIFF
--- a/bin/bc
+++ b/bin/bc
@@ -16,13 +16,6 @@ $yysccsid = "@(#)yaccpar 1.8 (Berkeley) 01/20/91 (Perl 2.0 12/31/92)";
 #define YYBYACC 1
 #line 49 "bc.y"
 
-
-;# I don't use BigFloat any more because they lack operators such as **,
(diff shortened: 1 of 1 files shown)


briandfoy:

briandfoy (approved):

PR #154, opened 2023-04-14
od: zero-fill input to required byte length

mknos:
* I noticed that when the number of input bytes is odd, the output of od -x is incorrect
* This was because unpack('S', $data) returns nothing if $data is 1 byte
* Standard od zero-fills the 16 bit number
* The do-remainder-bytes logic is incorrect (tangled web of pack, unpack and substr)
* float() and long() require input to be zero-filled 32-bits
* udecimal(), decimal(), octal2() and hex() require input to be zero-filled 16-bits
* Patch tested on amd64 against GNU od

Original output which uncovered this:
$ perl hexdump test
00000000  c6 f3                                             |..|
$ od -j 1 -x test # gnu od
0000001 00f3
0000002
$ perl od -j 1 -x test
00000001 0001
00000002

Also without -j:
$ xxd in2 
00000000: d8ab 7a                                  ..z
$ perl od -x in2
00000000 abd8 0000
00000003

### DIFF
--- a/bin/od
+++ b/bin/od
@@ -118,7 +118,7 @@ close $fh;
 exit 0;
 
 sub octal1 {
-    $upformat = 'c*'; # for -b
+    $upformat = 'C*'; # for -b
     $pffmt = '%.3o ';
     @arr = unpack($upformat,$data);
     $strfmt = $pffmt x $len;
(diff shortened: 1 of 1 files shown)


briandfoy:

briandfoy (approved):

Issue #161, opened 2023-06-14
units script produces incorrect results for definitions that include a fractional factor


jgpuckering:
The units script incorrectly handles fractional multipliers in both entry and in definitions.  For example, the documentation (in the DESCRIPTION section) offers this usage example:

```
    You have: 1|2 inch
    You want: cm
        * 1.27
        / 0.78740157
```

Those numbers are correct.  However, the actual output when you enter this example is:

```
You have: 1|2 inch
You want: cm
        * 0.635
        / 1.5748
```

Note that the values are squares of the correct values.  Upon investigation I discovered that the lex subroutine is capturing fractions twice, causing the fraction to be multiplied by itself during reduction.  This problem also occurs when the definition includes a fraction, not just on entry of a fractional unit.  For example, the conversion from year to month (where month is defined in the DATA section as "1|12 year") yields this result:

```
You have: month
You want: year
        * 0.00694444
        / 144

```

The problem traces to this bit of code in the lex subroutine:

 ```
 my $N = '(?:\d+\.\d+|\d+|\.\d+)(?:[eE][-+]?\d+)?';

  my @t = split /(
                   (?: \*.\* | !.! ) # Special 'new unit' symbol
                |  [()*-]                    # Symbol
                |  \s*(?:\/|\bper\b)\s*      # Division
                |  ($N\|$N)                  # Fraction
                |  $N                        # Decimal number
                |  \d+                       # Integer
                |  [A-Za-z_][A-Za-z_.]*      # identifier
                |  \s+                       # White space
                )/ox, $s;

```

Note that the $N regex includes ?: to disable group capturing.  However, this does not carry through to the Fraction alternative within the split pattern.  Changing the subpattern to (?:$N|$N) seems to completely resolve the issue.

I will attempt to clone the repo, fix this bug, and then push the fix.

briandfoy:
Thanks for looking into this.

Issue #162, opened 2023-06-14
script "units" generates error messages even before asking for input


jgpuckering:
_I submitted this recently to RT Bug Tracker (see ticket 148683) not noticing the "Preferred bug tracker" heading.  So I'm resubmitting it here._

---------------------------

The units utility contains a table of unit definitions in the DATA section that includes temperature units.  Unfortunately the formulae for these units include subtraction, and subtraction is not supported by the units parser.  Moreover, some definitions are incomplete (lacking a scaling factor).

Here's what you get when you run the version of units included in PPT 1.034:

Show quoted text

```
Parse error: Syntax error.  Skipping.
Parse error: Unknown unit `F'.  Skipping.
Parse error: Syntax error.  Skipping.
Parse error: Unknown unit `C'.  Skipping.
Parse error: Unknown unit `degC'.  Skipping.
Parse error: Unknown unit `kelvin'.  Skipping.
Parse error: Unknown unit `degC'.  Skipping.
```

I took a run at fixing this by trying to extend the parser to include affine conversions, but after several attempts I gave up on that approach.  

One obstacle I ran into is that the current logic relies a great deal on using unity ("1") as the identity operand during reductions.  That doesn't work with affine transformations like temperature, which need scaling (multiplication and division) as well as translation (addition and subtraction).  Fixing it seemed to involve a lot of code rework, including adding at least two more state transitions to a parser that is a struggle to understand with twelve.  (Well, for me anyway!)  And with no test script in the distribution the risk of regression seemed too high.

Another obstacle is that the output is displayed as a multiplication factor and a division factor -- thus presuming all conversions involve linear scaling.  If I ask to convert "cm" to "in" I get * 0.393701 and / 2.54.  If I change that to "100 cm", I get * 39.3701 and / 0.0254.  But if I'm converting 100 degrees Celsius to Fahrenheit, I want to see a result of 212 F, not scaling factors that only make sense for a specific temperature quantity.   Moreover, the default quantity is 1, whereas 0 is more sensible for temperature.  And if you enter a zero multiplier (e.g. "0 K") you get a PARSE_ERROR.

[...]

PR #163, opened 2023-06-14
Fixes to the units utility addressing issues #161 (fractions) and #162 (temperature units)

jgpuckering:
This sequence of commits addresses the above-noted issues.  From earliest to latest, here's what they include:

1. Replace backticks in six places where they were (likely inadvertently) paired with apostrophes.  Now they match.
2. Fix issue #161 "units script produces incorrect results for definitions that include a fractional factor" by adding non-capturing group directive to the Fraction pattern in sub lex.
3. Add a test script and restructure units to support (a) calling from a test script, (b) returning as a hash when testing; and (c) supporting a non-interactive mode when two units are given as command-line arguments.
4. Implement support for temperature conversions; add a new debug mode (T for temperature); revise temperature definitions in the DATA table; update the POD; add test cases for temperature units.
5. In sub lex, eliminate a redundant subpattern in the "Special 'new unit' symbol" split subpattern.
6. Document the optional units command line arguments in the synopsis.

I've run the test script my local environment (Windows 11) and they pass in both Windows and Ubuntu.  Interactive testing worked for me too.

### DIFF
--- a/bin/units
+++ b/bin/units
@@ -80,11 +80,45 @@ sub run {
   my( $class, @args ) = @_; local @ARGV;
 
   my $args = $class->process_args( @args );
+
   $class->read_unittab( $args->{unittabs}[0] );
 
-  $class->input_loop;
+  if (@{ $args->{args} }) {
+    my ($have, $want) = @{ $args->{args} };
+    my $have_hr = $class->unit_have($have);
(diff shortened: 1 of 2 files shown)


briandfoy:
Looks good. Thanks,

PR #164, opened 2023-06-15
Add metadata POD to base64 and hexdump

jgpuckering:
These were the only two file actually missing a metadata section.

### DIFF
--- a/bin/base64
+++ b/bin/base64
@@ -1,5 +1,19 @@
 #!/usr/bin/perl
 
+=begin metadata
+
+Name: base64
+Description: encode and decode base64 data
+Author: Michael Mikonos
+License: perl
+
+=end metadata
+
+=cut
+
+use strict;
+use warnings;
+
 use Getopt::Std qw(getopts);
 use MIME::Base64 qw(decode_base64 encode_base64);
 
--- a/bin/hexdump
+++ b/bin/hexdump
@@ -1,5 +1,19 @@
 #!/usr/bin/perl
 
+=begin metadata
+
+Name: hexdump
+Description: print input as hexadecimal
+Author: Michael Mikonos
+License: 
+
+=end metadata
+
+=cut
+
+use strict;
+use warnings;
+
 use Getopt::Std qw(getopts);
 
 my (%opt, @chars, @cesc, $adr, $nread, $curf, $skip);


bin/hexdump
```diff
@@ -1,5 +1,19 @@
 #!/usr/bin/perl
 
+=begin metadata
+
+Name: hexdump
+Description: print input as hexadecimal
+Author: Michael Mikonos
+License:
```
briandfoy, line 8: This license can be artistic2 by default since it's a new program, but Michael can also choose one of the others from the LICENSE files in the distro.

bin/hexdump
```diff
@@ -1,5 +1,19 @@
...
+Name: hexdump
+Description: print input as hexadecimal
+Author: Michael Mikonos
+License: 
+
+=end metadata
+
+=cut
+
+use strict;
```
briandfoy, line 14: This is fine for now, but this is a change outside the scope of the pull request. When something like this is included, I have to do extra work to check that it's okay. Without this, I could have immediately approved and merged the metadata change.

It's okay to add strict and warnings, but that's part of a larger issue that needs more care. Mainly, is the program warnings free for every version of Perl since v5.008? If they aren't, we also need to fix that to have warnings enabled.

bin/base64
```diff
@@ -1,5 +1,19 @@
 #!/usr/bin/perl
 
+=begin metadata
+
+Name: base64
+Description: encode and decode base64 data
+Author: Michael Mikonos
+License: perl
```
briandfoy, line 8: Change this to artistic2. Or, ask Michael what he wants.

We want to avoid the GPL for new things because that causes problems for some people and it doesn't really benefit this project. (I know you said you were a license compliance guy in a former life).

Some other programs in this repo use GPL, but that's what they were originally so that's how it is. I'd be inclined to replace those with new versions just to reduce the license count.

briandfoy (changes requested):
There are a few license things to clarify

jgpuckering:
Bear with me as I’m still getting used to GitHub’s collaboration environment.

I forgot I had added use strict and warnings last night, before your message about digestible pull requests.  I should have removed those.  Sorry about that.  And thanks for the reminder about the need to deal with feature warnings across Perl versions.  I understand the issue.  There should be a tool for that …or a set of policies that warn of about such issues.

Regarding the licenses, I gather it is my action to reach out to Michael.  Correct?

briandfoy:
You can ask Micheal about the licenses (@mknos to tag him) or I can. Maybe he'll pop in and tell us.

Everything else: no worries. It might even be easier to close this PR and redo it. Whatever works for you.

jgpuckering:
> You can ask Micheal about the licenses (@mknos to tag him) or I can. Maybe he'll pop in and tell us.
> 
> Everything else: no worries. It might even be easier to close this PR and redo it. Whatever works for you.

Micheal (@mknos) hasn't provided an email address on his GitHub profile page.  Do you have a way to contact him?

PR #180, opened 2023-07-05
cp: copy many files at once

mknos:
* Multiple source files can be given before a destination
* Some of what process_arguments() returns was being thrown away
* For case where system() is called, cp command can handle all files at once
* For case where File::Copy::copy() is called, introduce a loop because it can't handle a list of source files
* Exit early if insufficient file arguments are provided: 0 (cp), 1 (cp srcfile)


briandfoy (approved):
Good catch, thanks,

PR #191, opened 2023-07-06
cmp: directory arguments

mknos:
* cmp command compares two files which may be binary but may not be a directory
* I was getting a success exit code on my system when running: perl cmp .. file.txt
* Check for directory before open() is called
* cmp exits 0 if the compared files match and 1 if they don't, so exit code 2 is designated for actual errors

### DIFF
--- a/bin/cmp
+++ b/bin/cmp
@@ -42,6 +42,8 @@ END {
     $? = 1 if $? == 255;  # from die
 }
 
+$SIG{__DIE__} = sub { warn @_; exit 2 };
+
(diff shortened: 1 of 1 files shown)


briandfoy (approved):
I'm going to fix this up myself, but let's stop using die handlers like this. We might want to die for some other reason and exit with some other value. I'd rather see warn and exit.

briandfoy:
I've fixed up this in 209ca83223c200453d30c836bcec64e3a0607ca2, and added [PROGRAMMING_STYLE.md].

PR #198, opened 2023-07-10
rm: ignore directory arg unless -r is set

mknos:
* When I compare rm with and without -r, rmdir() appears to be called without -r
* I run... mkdir dir0; touch dir0/file0; perl rm dir0 # rmdir() fails becuse dir0 is not empty
* Standard rm should not remove a directory without the -r flag
* GNU rm and OpenBSD rm print a warning to stderr to indicate the directory argument is being ignored; follow this
* Global variable $arg is unused

### DIFF
--- a/bin/rm
+++ b/bin/rm
@@ -46,7 +46,6 @@ use Getopt::Std;
 
 # Command line arguments, and other variables.
 use vars qw( $opt_i $opt_f $opt_r $opt_R $opt_P );
-my $arg = 0;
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy (approved):

briandfoy:
This program is a good candidate for conversion to GetOpt::Std, which could then use `-d` too.

Thanks,

PR #199, opened 2023-07-10
od: simplify input loop

mknos:
* The input loop reads 16 bytes which are formatted as one line of output
* Move the formatting determination out of the loop because the options are only set once
* Introduce $fmt, referring to the formatter sub we need
* Calling formatter effectively sets @arr and $strfmt, both of which are passed to printf()


briandfoy (approved):
Oh, this looks much nicer than the original.

PR #200, opened 2023-07-11
rm: exit code 1 on error

mknos:
* Exit code 1 indicates an error occurred when processing one or more arguments
* An error processing one argument doesn't prevent the next argument from processing
* Flag error (printing $! if appropriate) for these cases:
1) expected file, got directory
2) rmdir() failed on a directory (in -r mode)
3) opendir() failed (in -r mode), e.g. permission denied
4) unlink() failed, e.g. file does not exist
5) expected an argument, got zero

* closedir() isn't needed if opendir() failed
* While here remove -P flag for "wiping" a file
1) not part of standard rm command
2) implementation assumes writing over a file in place will effectively wipe content on disk
3) implementation assumes the file being wiped can fit in memory (creates a string the same size as file)
4) physical destruction is *always* better

### DIFF
--- a/bin/rm
+++ b/bin/rm
@@ -45,17 +45,20 @@ use strict;
 use Getopt::Std;
 
 # Command line arguments, and other variables.
-use vars qw( $opt_i $opt_f $opt_r $opt_R $opt_P );
+use vars qw( $opt_i $opt_f $opt_r $opt_R );
 
-# Get the options.
-getOptions();
-
-#
-# Process each file named on the command line.
+getopts('ifrR') or showUsage();
+unless (@ARGV)
+{
+    warn "$0: missing argument\n";
+    exit 1;
+}
+my $Err = 0;
 for my $arg ( @ARGV )
 {
     processFile( $arg );
 }
+exit($Err ? 1 : 0);
 
 #
 #  Attempt to process each file / directory named on the command line.
@@ -68,6 +71,7 @@ sub processFile()
 	if (!$opt_r && !$opt_R)
 	{
 	    warn "$0: cannot remove '$fileName': is a directory\n";
+	    $Err++;
 	    return;
 	}
 	removeDirectory( $fileName );
@@ -87,8 +91,8 @@ sub removeDirectory( )
 
     unless (opendir(DIR, $dirName))
     {
-	warn "Can't open $dirName\n";
-	closedir(DIR);
+	warn "$0: cannot open '$dirName': $!\n";
+	$Err++;
 	return;
     }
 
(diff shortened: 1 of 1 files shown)


briandfoy:
I see this, but give me a couple days to look through it since this one is a bit more involved than the other PRs you've had.

briandfoy:
I agree on `-P`, but maybe we should just make it a no-op instead of removing it. On macOS:

```
-P      This flag has no effect.  It is kept only for backwards
             compatibility with 4.4BSD-Lite2.
```

bin/rm
```diff
@@ -45,17 +45,20 @@ use strict;
 use Getopt::Std;
 
 # Command line arguments, and other variables.
-use vars qw( $opt_i $opt_f $opt_r $opt_R $opt_P );
```
briandfoy, line 48: Let's make $opt_P a no-op instead. Maybe issue a warning.

bin/rm
```diff
@@ -45,17 +45,20 @@ use strict;
...
-
-#
-# Process each file named on the command line.
+getopts('ifrR') or showUsage();
+unless (@ARGV)
+{
+    warn "$0: missing argument\n";
+    exit 1;
+}
+my $Err = 0;
```
briandfoy, line 56: Maybe the return value of processFile should decide if something gets added to $Err? I'd rather have that then subroutines accessing variables outside their scope.

bin/rm
```diff
@@ -147,84 +155,20 @@ sub removeFile( $fileName )
...
-    }
-
     # Delete the file.
-    unlink( $fileName );
-}
-
-#
-# Overwrite the file specified, first with x00, the xFF, then x00
-sub overWriteFile( )
-{
```
briandfoy, line 163: I'd turn overWriteFile into a warning and a return.

briandfoy (changes requested):

briandfoy:
So, on alpine linux, I get exit of 1 with no arguments, but on macOS and freebsd I get 64. This is not a big deal.

I recently made https://github.com/briandfoy/PerlPowerTools/blob/master/PROGRAMMING_STYLE.md , which has a few things to say about exit values. You don't need to do all that stuff for this PR, but I'd gradually like to move things toward modulinos and a common structure among programs.

mknos:
* NetBSD, OpenBSD, FreeBSD (I think Mac follows this), HP-UX  manuals are consistent: exit code >0 means error
* This patch doesn't distinguish how many arguments failed to process, or the processing error that occurred; exit code 1 is used for everything
* This could be changed later but this patch is at least compliant with standard

briandfoy (approved):

briandfoy:
I merged this, but I'm going to add additional changes. I started working on this program and it got a bit out of hand as I fixed the -i and -f issue.

Thanks,

PR #201, opened 2023-07-12
Add unlink command

mknos:
* unlink is a trivial rm that processes a single file argument and has no option for processing directories
* The command is provided by FreeBSD, DragonFlyBSD and Linux
* We can provide it here too (mainly for the benefit of shell scripts)

### DIFF
--- a/bin/unlink
+++ b/bin/unlink
@@ -1,0 +1,77 @@
+#!/usr/bin/perl
+
+=begin metadata
+
+Name: unlink
+Description: simpler than rm
+Author: Michael Mikonos
+License:
+
+=end metadata
+
+=cut
+
+use strict;
+
+use File::Basename qw(basename);
+use Getopt::Std qw(getopts);
+
+use constant EX_SUCCESS => 0;
+use constant EX_FAILURE => 1;
+
+my $Program = basename($0);
+
+unless (getopts('')) {
+    warn "usage: $Program FILE\n";
+    exit EX_FAILURE;
+}
+unless (@ARGV) {
+    warn "$Program: missing operand\n";
+    exit EX_FAILURE;
+}
+my $file = shift;
+if (@ARGV) {
+    warn "$Program: extra operand: '$ARGV[0]'\n";
+    exit EX_FAILURE;
+}
+if (-d $file) {
+    warn "$Program: cannot unlink '$file': is a directory\n";
+    exit EX_FAILURE;
+}
+unless (unlink $file) {
+    warn "$Program: cannot unlink '$file': $!\n";
+    exit EX_FAILURE;
+}
+exit EX_SUCCESS;
+
+__END__
+
+=pod
+
+=head1 NAME
+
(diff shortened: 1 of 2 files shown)


bin/unlink
```diff
@@ -0,0 +1,77 @@
...
+
+use File::Basename qw(basename);
+use Getopt::Std qw(getopts);
+
+use constant EX_SUCCESS => 0;
+use constant EX_FAILURE => 1;
+
+my $Program = basename($0);
+
+unless (getopts('')) {
```
briandfoy, line 24: What's going on here? Are you simply failing if there are any options?

bin/unlink
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/perl
+
+=begin metadata
+
+Name: unlink
+Description: simpler than rm
+Author: Michael Mikonos
+License:
```
briandfoy, line 8: Choose a license. I suggest artistic2

bin/unlink
```diff
@@ -0,0 +1,77 @@
...
+
+=head1 AUTHOR
+
+Written by Michael Mikonos.
+
+=head1 COPYRIGHT
+
+Copyright (c) 2023 Michael Mikonos.
+
+This program is free software; you can redistribute it and/or modify it under
```
briandfoy, line 76: This needs to match whatever you choose in the license metadata

bin/unlink
briandfoy: I'm curious how this would deal with unlinking a file that starts with a '-'. Can you check that?

bin/unlink
```diff
@@ -0,0 +1,77 @@
...
+}
+unless (unlink $file) {
+    warn "$Program: cannot unlink '$file': $!\n";
+    exit EX_FAILURE;
+}
+exit EX_SUCCESS;
+
+__END__
+
+=pod
```
briandfoy, line 49: Add =encoding utf8 to start pod. This is simply something I'm trying to make consistent everywhere.

briandfoy (changes requested):

bin/unlink
mknos: I call getopt() with no options...
a) so "--" is not treated as a file ("unlink -- FILE" is ok but "unlink --" is incomplete)
b) so arguments like "-f" are treated as an option (unknown option error + usage string is printed)
This is how the GNU version of unlink behaves.

The file "-" is allowed and doesn't require special escaping.
The files "-f" and "--" would normally be given as absolute path or relative to "." ("./-f" and "./--").

briandfoy (approved):

briandfoy:
This looks good.

There's an additional task that we need for new programs. In the gh-pages branch, update index.html and */index.html to include the new program name. This is in the CONTRIBUTING.md doc, but I've done it already so there's nothing to do here. That would be a pull request on its own since it has to target the gh-pages branch instead of master.

PR #202, opened 2023-07-13
base64: fail early for directory argument

mknos:
* I am still reading through how various commands are treating directories
* Future changes: possibly add special handling for files containing garbage (non-base64 characters)

### DIFF
--- a/bin/base64
+++ b/bin/base64
@@ -16,33 +16,45 @@ License: artistic2
 use strict;
 use warnings;
 
+use constant EX_SUCCESS => 0;
+use constant EX_FAILURE => 1;
+
+use File::Basename qw(basename);
 use Getopt::Std qw(getopts);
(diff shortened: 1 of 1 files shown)


bin/base64
```diff
@@ -38,6 +40,7 @@ if (defined $opt{'o'}) {
     $out = *STDOUT;
 }
 if (defined $ARGV[0] && $ARGV[0] ne '-') {
+    die "$0: '$ARGV[0]' is a directory\n" if (-d $ARGV[0]);
```
briandfoy, line 43: Instead of die-ing, warn and exit with an appropriate exit value.

briandfoy: Also, instead of using $0, use the basename of $0. I have some examples in PROGRAMMING_STYLE.md

briandfoy (changes requested):

briandfoy (approved):

PR #203, opened 2023-07-14
od: add -N option: limit input bytes

mknos:
* Standard od has -N which takes a numeric argument that is interpreted as decimal, hex or octal
* Initially add support for only decimal argument; this is the default interpretation in the standard
* -N 0 is allowed
* -j can be used with -N; -j bytes are skipped first before -N bytes are read
* Raise error for skipping past end of file with -j
* Avoid reading more bytes than -N by switching input loop to single byte read (prevents hanging waiting for input when -N <16 and reading from stdin connected to terminal)
* Tested against GNU od

### DIFF
--- a/bin/od
+++ b/bin/od
@@ -109,6 +109,11 @@ if ($offset1) {
 	undef $data;
     }
 }
+if (defined($lim) && $lim == 0) {
+    printf("%.8$radix\n", $offset1);
+    close $fh;
+    exit EX_SUCCESS;
+}
 
 $opt_o = 1 if ! ($opt_b || $opt_c || $opt_d || $opt_f || $opt_i ||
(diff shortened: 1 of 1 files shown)


bin/od
```diff
@@ -217,7 +266,7 @@ sub diffdata {
 }
 
 sub help {
-    print "usage: od [-bcdfiloxv] [-A radix] filename\n";
+    print "usage: od [-bcdfiloxv] [-A radix] [-j skip_bytes] [-N limit_bytes] filename\n";
     exit 1;
```
briandfoy, line 270: while we are here, we might as well update this exit too.

briandfoy (changes requested):

briandfoy (approved):

PR #204, opened 2023-07-16
od: validate radix argument to -A

mknos:
* Standard od allows "d", "o", "x" and "n" for radix value for Decimal, Octal, Hex and None
* I noticed that this version doesn't validate -A before passing it to printf()
* Offset is not supposed to be printed for -A n (including line prefix and for final summary line)
* The default value of -A is "o"

### DIFF
--- a/bin/od
+++ b/bin/od
@@ -69,7 +69,7 @@ my $Program = basename($0);
 
 getopts('A:bcdfij:lN:ovx') or help();
 if (defined $opt_A) {
-    if ($opt_A !~ m/^[doxn]$/) {
+    if ($opt_A !~ m/\A[doxn]\z/) {
 	warn "$Program: unexpected radix: '$opt_A'\n";
 	exit EX_FAILURE;
     }
--- a/bin/od
+++ b/bin/od
@@ -68,7 +68,19 @@ $lastline = '';
 my $Program = basename($0);
 
 getopts('A:bcdfij:lN:ovx') or help();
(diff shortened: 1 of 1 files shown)


bin/od
```diff
@@ -68,7 +68,19 @@ $lastline = '';
 my $Program = basename($0);
 
 getopts('A:bcdfij:lN:ovx') or help();
-    defined $opt_A ? ($radix = $opt_A) : ($radix = 'o');
+if (defined $opt_A) {
+    if ($opt_A !~ m/^[doxn]$/) {
```
briandfoy, line 72: I'd change that to /\A[doxn]\z/. The ^ and $ can allow more that what you are trying to validate, and change their effect with certain re module flags.

bin/od
```diff
@@ -110,7 +122,7 @@ if ($offset1) {
     }
 }
 if (defined($lim) && $lim == 0) {
-    printf("%.8$radix\n", $offset1);
+    printf("%.8$radix\n", $offset1) if $radix;
```
briandfoy, line 125: Is there ever a case where $radix won't have a value?

briandfoy (changes requested):

bin/od
mknos: With this patch, $radix won't be assigned a value if $opt_A is "n"

briandfoy: ah, my fault. I read that branch above as `eq` and not `ne`.

briandfoy (approved):

PR #206, opened 2023-07-18
Bcd import

mknos:
* This project has the bsd games "morse" and "ppt" but not "bcd"
* New script is a translation of (most of) the OpenBSD version from the C code
* The output is equivalent to bcd on my OpenBSD system
* The bsdgames package on my ubuntu system doesn't have the bcd -l flag; that's an OpenBSD thing
* The output is not entirely equivalent to bcd on my ubuntu system. Refer to latest commit 5 years ago:
https://cvsweb.openbsd.org/src/games/bcd/bcd.c
* Todo: The decode() function (-d flag) is not added yet

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -9,6 +9,7 @@ bin/banner
 bin/base64
 bin/basename
 bin/bc
+bin/bcd
 bin/cal
 bin/cat
 bin/chgrp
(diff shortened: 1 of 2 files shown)


briandfoy (approved):

briandfoy:
@mknos Would you like to try a PR against the gh-pages branch to add your new program to the various index.html pages?

The gh-pages branch is a disconnected branch that is never merged, but it's the one the GitHub uses to build the website.

PR #208, opened 2023-07-20
od: -c flag versus non-printable characters

mknos:
* I made a test file with bytes from 0-255
* I ran "perl od -c tst" and saw warnings: Wide character in printf at od line 174.
* Standard od does not attempt to print non-printable ascii characters; instead, the character is printed as 3 octal digits
* If a c-escape is available for a non-printable character, use that instead
* unpack format should be "C" not "c" to treat characters as unsigned
* The warning appears to be present in older commits, so I guess I didn't break this recently


briandfoy (approved):
Yeah, I thought that big lookup table was weird when I last looked at this, but I didn't pursue it.

PR #211, opened 2023-07-21
ed: strict.pm

mknos:
* Having strict is desirable for maintainability, and many commands in this project already do this
* Delete unreachable code: out-of-order address check happens before calling command-specific functions, e.g. edPrint()
* Delete unused variable $NumLines in edSubstitute()
* Global variables are not being renamed, but tidy some of their comments
* Standard ed allows -s to silence the byte counts, but this version doesn't have it yet

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -71,37 +71,37 @@ License: gpl
 #        - Add BSD/GNU extentions
 #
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):
Good work!

PR #213, opened 2023-07-24
ed: join command (j) compatibility

mknos:
* Initial implementation of edJoin() didn't allow 1-address ("1j") or 0-address ("j") variants, only 2-address ("1,2j")
* Follow GNU ed and treat 1-address join as nop (does not raise an error or mark buffer as modified)
* Single line ranges were already treated as nop, e.g. "1,1j"
* For bare "j", join current line with subsequent line (error if current line is the final line)
* j command doesn't allow extra arguments, only addresses
* Also, edPrint() and edJoin() should not accept the special address 0 (accepted by some ed commands)

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -282,6 +282,10 @@ sub edPrint {
     $adrs[0] = $CurrentLineNum unless (defined($adrs[0]));
     $adrs[1] = $adrs[0] unless (defined($adrs[1]));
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):
Thanks again for the good work. Your issue comments really make it easy to review these things. 

Also, at some point, maybe you can just write ed altogether :)

PR #218, opened 2023-07-28
ed: add list (l) command

mknos:
* This feature was mentioned in a Todo comment
* "l is a variant of the "p" (print) command so handle this in edPrint()
* Non-printable characters are printed as "\" followed by three octal digits, with no extra space on output
* Some common c-escapes are printed except for \n
* GNU ed and OpenBSD ed don't print the c-escape \0 but display \000 instead
* Extra1: Make the if-elsif block treat $command consistently by using eq
* Extra2: bump version

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -57,7 +57,6 @@ License: gpl
 #        - Implement the following commands from the v7 docs:
 #                g - global command
 #                k - mark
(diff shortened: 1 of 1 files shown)


briandfoy (approved):
Nice work on changing all those regexes to `eq`

PR #221, opened 2023-07-31
ed: fix range-specific write

mknos:
* Command suffix was already saved into $args[0] so inventing $commandsuf was a bad idea (remove it)
* Detect whitespace after $command character correctly by making it a capture variable in command regex
* "w" command has 3 different usages so handle them all correctly
* usage1: no address (w filename): write entire buffer to file
* usage2: one address: (123w filename): write single line from buffer to file
* usage3: address range: (12,13w filename): write only the lines in inclusive range to file
* style: declare filehandle as a "my" variable in edWrite()


briandfoy:
I see this, but I want to go through it a bit more carefully to see what you did. I'll try to get to that soon.

briandfoy (approved):

PR #222, opened 2023-08-15
rev: accept "?" as a file

mknos:
* Argument parsing in rev treated "?" as an indication to print usage, but this is not standard behaviour
* GNU rev and OpenBSD rev accept a file called "?" and we could accept it here
* I tested it with: echo hello > \? && perl rev \?

### DIFF
--- a/bin/rev
+++ b/bin/rev
@@ -18,7 +18,7 @@ $|++;
 
 my ($VERSION) = '1.3';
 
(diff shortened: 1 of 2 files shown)


briandfoy (changes requested):
There's more more change you need. Remove the `?` test in *t/rev/rev.t*. It's in the last subtest. That should bet the CI checks to pass.

And, while you are there, can you rename that subtest `help` instead of `version`? It's obvious that I copied and pasted there.

mknos:
Thanks for the suggestion. Unit test updated.

briandfoy (approved):

PR #223, opened 2023-08-24
cat: exit(1) for bad arguments

mknos:
* Previously on my system directory arguments would silently be ignored instead of flagging an error
* Make $options more pretty
* Declare $Program and exit code constants as done in other scripts
* Explicitly check for directory before open() as done in other scripts
* Error exit if one or more files cannot be opened, or is a directory
* Lone '-' argument is still handled for stdin
* The <> loop over all input files becomes explicit do_file() call per argument
* case1: `perl cat F1 .` will print F1, warn about directory '.' then exit(1)
* case2: `perl cat F1 F2` will warn about F1 not existing, proceed to print F2 then exit(1)

### DIFF
--- a/bin/cat
+++ b/bin/cat
@@ -11,28 +11,30 @@ License: perl
 
 =cut
 
-
 use strict;
-use Getopt::Std;
+
+use File::Basename qw(basename);
+use Getopt::Std qw(getopts);
+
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
Very nice. Thanks,

PR #226, opened 2023-09-02
fold: continue on error

mknos:
* Match GNU fold: If an argument can't be opened, proceed to the next argument instead of exiting
* When arguments are finished, exit with error code if we failed for one or more argument
* Avoid calling open() if argument is a directory
* Fix handling of special filename '-' (for both "fold -" and bare "fold")
* Align usage string with SYNOPSIS: file argument is optional
* Prefix warning messages with program name

### DIFF
--- a/bin/fold
+++ b/bin/fold
@@ -27,6 +27,13 @@ License: perl
 use strict;
 use locale;             # for what a space is.
 
+use File::Basename qw(basename);
+
+use constant EX_SUCCESS => 0;
+use constant EX_FAILURE => 1;
+
+my $Program = basename($0);
+
 my(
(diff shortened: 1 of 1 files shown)


bin/fold
```diff
@@ -87,11 +90,11 @@ if ($Space_Break && $Width < $Tabstop) {
 }
 
 unshift(@ARGV, '-') unless @ARGV;
+my $err = 0;
 for (@ARGV) {
-    fold_file($_);
+    $err |= fold_file($_);
```
briandfoy, line 95: I typically increment instead of toggling, but this is fine,. Haven't seen this in awhile. Fond memories. :)

briandfoy (approved):
Nice job. Thanks,

PR #227, opened 2023-09-04
head versus directories

mknos:
* No need to validate $count if the default value is used
* Raise an error for -n 0
* Reject directory arguments
* Proceed to next argument if an argument failed to process; exit with failure code when arguments are done
* compat1: print empty line before title line of next file (except for 1st file)
* compat2: don't print a space afer "<==" in title line

### DIFF
--- a/bin/head
+++ b/bin/head
@@ -31,7 +31,7 @@ unless (getopts('n:', \%opt)) {
 my $count;
 if (defined $opt{'n'}) {
     $count = $opt{'n'};
-    if ($count =~ m/\D/) {
+    if ($count =~ m/\D/a) {
         warn "$Program: invalid number '$count'\n";
         exit EX_FAILURE;
     }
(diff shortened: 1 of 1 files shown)


bin/head
```diff
@@ -13,32 +13,66 @@ License: perl
...
-die "invalid number `$count'\n" if $count =~ /\D/;
+my %opt;
+unless (getopts('n:', \%opt)) {
+    warn "usage: $Program [-n count] [file ...]\n";
+    exit EX_FAILURE;
+}
+my $count;
+if (defined $opt{'n'}) {
+    $count = $opt{'n'};
+    if ($count =~ m/\D/) {
```
briandfoy, line 34: I'd write this as m/\D/a so it uses the ASCII version of a digit.

briandfoy (changes requested):

PR #230, opened 2023-09-07
nl: exit(1) for bad file argument

mknos:
* We already support multiple file arguments, similar to GNU nl ('-' is stdin)
* Follow GNU nl further: exit with error code if a file argument fails to process
* Previously no error was printed for a directory argument
* Follow OpenBSD nl: allow -i0 (GNU nl does not allow it and -i0 was previously treated as -i1)
* Argument to -i is a signed int and may be written with '+' or '-' prefix
* We already support negative -i value (similar to OpenBSD nl) but GNU nl does not allow it
* Bump version

### DIFF
--- a/bin/nl
+++ b/bin/nl
@@ -29,7 +29,7 @@ use File::Basename qw(basename);
 use constant EX_SUCCESS => 0;
 use constant EX_FAILURE => 1;
 
-our $VERSION = '1.01';
+our $VERSION = '1.2';
 my $program  = basename($0);
 my $usage    = <<EOF;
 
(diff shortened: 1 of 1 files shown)


briandfoy:
I see this but I want to look at it more closely, so I'll get to that soon.

mknos:
The other way to proceed would be to simplify the code and switch over to the BSD behaviour.
NetBSD, OpenBSD and FreeBSD (probably mac osx by extension) support only a single file as input.

briandfoy:
Switching to BSD behavior is fine. That's always been Tom's intent for Perl PowerTools. I certainly don't think that we should try to merge two different implementations. We want to be exactly like at least on of them and that one is the BSD branch when possible.

briandfoy (approved):

PR #235, opened 2023-09-09
bc: integer precision problem with -b mode

mknos:
* Previous bignum code was not correct because digit information was lost before conversion to BigFloat object
* Digits after the decimal point didn't seem to be affected though
* This version works better for me

bc -y # old
123384888888888888888888888888888888888888888888888812 123384888888889000000000000000000000000000000000000000

1.234448888888888888888888888888888888888888888888887 1.234448888888888888888888888888888888888888888888887

bc -y # new
123384888888888888888888888888888888888888888888888812 123384888888888888888888888888888888888888888888888812

### DIFF
--- a/bin/bc
+++ b/bin/bc
@@ -2164,11 +2164,10 @@ sub yylex
 
           # number, is it integer or float?
 	  if ($line =~ s/^(\d+)//) {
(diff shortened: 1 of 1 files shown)


briandfoy (approved):
Nice change here.

PR #239, opened 2023-09-11
sum: more readable algorithms

mknos:
* Declare one buffer size constant used by all algorithms
* Standard sum command has no -h option for help; treating it as an unknown option still prints usage string
* In sum2() remove unused variable $i
* In sum1() and crc32() replace ord+substr pattern with unpacking $buf before loop
* My test file produces same sum for -o {0,1,2} as previous version

### DIFF
--- a/bin/sum
+++ b/bin/sum
@@ -18,11 +18,17 @@ License:
 
 use strict;
 use integer;
-use Getopt::Std;
-use vars qw($opt_h $opt_o);
 
-&help unless getopts('ho:');
-&help if ( $opt_h );
+use Getopt::Std qw(getopts);
+
+use vars qw($opt_o);
+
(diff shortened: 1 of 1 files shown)


bin/sum
```diff
@@ -195,12 +207,11 @@ sum - display file checksums and block counts
 
 B<sum>
 [ B<-o> I<0|1|2> ]
-[ B<-h> ]
 [ I<filename ...> ]
 
 =head1 DESCRIPTION
 
-sum outputs three space seperated values:  file CRC, file size, and
+sum outputs three space separated values:  file CRC, file size, and
```
briandfoy, line 214: This must be the most common typo. You've fixed this in more than a couple programs I think.

PR #243, opened 2023-09-13
rmdir: include filename in warning

mknos:
* Even in -p mode, GNU rmdir lists errors, but we didn't print them
* Because of truth check, we didn't remove the final 0/ component in "rmdir -p 0/0/0"
* 1/100 directory arguments might fail to remove and we didn't list the name in the error message
* Which files failed?

  rmdir: Not a directory
  rmdir: Not a directory
  rmdir: No such file or directory

* Patched version works for "-p 0/0/0" and prints a warning if one directory wasn't empty
* POD text mentioning "head" was possibly copied from another file

### DIFF
--- a/bin/rmdir
+++ b/bin/rmdir
@@ -13,60 +13,42 @@ License: perl
 
 
 use strict;
-use Getopt::Std;
+
+use File::Basename qw(basename dirname);
+use Getopt::Std qw(getopts);
+
+use constant EX_SUCCESS => 0;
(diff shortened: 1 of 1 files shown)


briandfoy (approved):
This is a good change, and the next step would be to match the error messages to the real rmdir. I get errors like `rmdir: foo/0: No such file or directory` instead of what we are doing now. That's not necessary for this to go in though.

Issue #174, opened 2023-06-21
bc: test if


briandfoy:

mknos:
$ cat bc.sh 
#!/bin/sh

BC='perl bc'

bc()
{
    code="$1"
    echo "$code" | $BC | fgrep -c ok
}

for tst in \
  'if (0 == 0) { print "ok"; }' \
  'if (1 != 0) { print "ok"; }' \
  'if (1 > 0) { print "ok"; }' \
  'if (0 < 1) { print "ok"; }' \
  'a=2; if (a == 2) { print "ok"; }' \
  'a=2; if (a == 2 && 1 == 1) { print "ok"; }' \
; do
    ok=$(bc "$tst")
    echo "SUCCESS=$ok -- $tst"
    if [ $ok != 1 ]; then
        echo ABORT
	exit 1
    fi
done
echo 'ALL OK'
exit 0

$ ./bc.sh  
SUCCESS=1 -- if (0 == 0) { print "ok"; }
SUCCESS=1 -- if (1 != 0) { print "ok"; }
SUCCESS=1 -- if (1 > 0) { print "ok"; }
SUCCESS=1 -- if (0 < 1) { print "ok"; }
SUCCESS=1 -- a=2; if (a == 2) { print "ok"; }
SUCCESS=0 -- a=2; if (a == 2 && 1 == 1) { print "ok"; }
ABORT

briandfoy:
Thanks, I'll look at this and integrate it and move on your PR.

mknos:
> SUCCESS=0 -- a=2; if (a == 2 && 1 == 1) { print "ok"; }

I didn't investigate why this was happening. But now I think it could be an order of operation problem because adding more parens makes it evaluate to true:

a=2; if ((a==2) && (1==1)) { print "ok"; }

PR #247, opened 2023-09-14
bc: don't print 0 for a false if-condition

mknos:
* perl bc evaluates false "if" as an expression and returns 0, not "null"

>>> if (0 == 1) { a = 1; } 
0

* This doesn't happen if the condition is true
* I observe that if $res==0, the executed statement result ($val) is printed
* In IF-statement code, $val is returned by exec_stmt() if the condition is true
* If the condition is false, $val defaults to zero but this causes an unwanted print later
* Changing the default to undef works for my simple test.bc script
* Reading the code, I see other examples of undef being pushed into ope_stack, so I think this will not corrupt the state
* Also remove unused variable $n found by perlcritic
* Does this look ok?

### DIFF
--- a/bin/bc
+++ b/bin/bc
@@ -2533,10 +2533,11 @@ sub exec_stmt
 # IF statement
 
      my $cond = pop @ope_stack;
-     $val = 0;
      if($cond) {
        ($return, $val) = exec_stmt($instr->[1]);
        push(@ope_stack, $val), last INSTR if $return;
+     } else {
+       $val = undef;
      }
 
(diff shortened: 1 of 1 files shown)


briandfoy:
Is this possibly related to the low-detail issue #174? A prior contributor noted that if might be doing the wrong thing but I hadn't looked into it. I know that there's no context in that issue, but if you have tried the various ways to use `if`, found that is was acting weird, and fixed it, I think I can close that issue too.

briandfoy (approved):

PR #252, opened 2023-09-18
colrm: add validation

mknos:
* Zero was allowed for column number, but it's not valid because we count from one---raise error like BSD version does
* Also, colrm allows negative value arguments; add validation to avoid this
* For ARGV==1, when arg1 > column count, output incorrectly gets another newline appended because the substring is the full string including newline
* For ARGV==2, "colrm 1 1" is supposed to remove only the 1st column but it did nothing
* Also, the substring could be outside the string for "colrm 1000 1000" because input was not validated

### DIFF
--- a/bin/colrm
+++ b/bin/colrm
@@ -14,31 +14,57 @@ License: perl
 
 use strict;
 
-$0 =~ s(.*/)();
-my $usage = "usage: $0 [startcol endcol]\n";
-my ($startcol, $endcol);
+use File::Basename qw(basename);
 
-# I could be more clever, but this will run faster
+my $Program = basename($0);
 
 if (@ARGV > 2) {
-	die $usage;
+	die "usage: $Program [startcol [endcol]]\n";
 } elsif (@ARGV == 0) {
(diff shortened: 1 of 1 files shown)


bin/colrm
```diff
@@ -14,31 +14,57 @@ License: perl
...
-my $usage = "usage: $0 [startcol endcol]\n";
-my ($startcol, $endcol);
+use File::Basename qw(basename);
 
-# I could be more clever, but this will run faster
+my $Program = basename($0);
 
 if (@ARGV > 2) {
-	die $usage;
+	die "usage: $Program [startcol [endcol]]\n";
```
briandfoy, line 22: let's change these to a `warn` and exit with the right value for the real colrm in the same situation

mknos: Ok. The netbsd and openbsd manuals didn't help much but the openbsd source does exit(1) within usage() function.

briandfoy: I usually go with whatever I discover is in $? when I try the real program :)

PR #259, opened 2023-09-23
cal: permit -j -y and -jy

mknos:
* Simplify options parsing 
* Previously -jy could be grouped, but not written as "-j -y" because options were expected to all be in $ARGV[0]
* num_args does not count options
* num_args == 0 means $month and $year are dynamically selected (month is not used if -y is set)
* num_args == 1 means $month is not used and $year comes from args
* num_args == 2 means $month and $year are both used and come from args (not compatible with -y)
* Remove non-standard -? option; usage is still printed
* Put POD after __END__
* Remove POD text referring to -n option which doesn't exist

### DIFF
--- a/bin/cal
+++ b/bin/cal
@@ -16,47 +16,25 @@ package
 	cal; # hide from PAUSE
 use strict;
 
+my %opts;
+while (defined($ARGV[0]) && substr($ARGV[0], 0, 1) eq '-') {
+	%opts = (%opts, get_flags(shift @ARGV));
+}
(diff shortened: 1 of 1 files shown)


bin/cal
```diff
@@ -16,47 +16,25 @@ package
 	cal; # hide from PAUSE
```
briandfoy, line 16: We could probably get rid of the special package too. Not a big deal or necessary for this.

briandfoy (approved):

PR #262, opened 2023-09-25
strings: better validation

mknos:
* Option -t expects a single letter, but this version accepted stupid things like -t bedding
* $offset_format is only used if $opt_t is set; default usage does not print offset
* Directory arguments were silently failing; now print an error message for them

### DIFF
--- a/bin/strings
+++ b/bin/strings
@@ -61,11 +61,18 @@ getopts( 'afon:t:' ) or die $usage;
 
 $opt_n ||= 4;
 $opt_t ||= 'o' if $opt_o;
-if ( $opt_t )
-{
-    $opt_t =~ /[dox]/ or die $usage;
-    $offset_format = "\%07$opt_t ";
+
+if ($opt_O) {
+    $opt_t = 'o';
+} elsif (defined $opt_t) {
+    my %EXPECT = (
+        'd' => 1,
+        'o' => 1,
+        'x' => 1,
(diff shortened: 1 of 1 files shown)


bin/strings
```diff
@@ -60,12 +60,17 @@ use Getopt::Std;
...
-    $offset_format = "\%07$opt_t ";
+if ($opt_o) {
+    $opt_t = 'o';
+} elsif (defined $opt_t) {
+    my %EXPECT = (
+        'd' => 1,
+        'o' => 1,
+        'x' => 1,
+    );
+    die($usage) unless $EXPECT{$opt_t};
```
briandfoy, line 71: Let's warn and exit with the right value here

bin/strings
```diff
@@ -76,6 +81,7 @@ $chunksize = 4096; # whatever
 
 for my $filename ( @ARGV )
 {
+    die("'$filename' is a directory\n") if (-d $filename);
```
briandfoy, line 84: Same here

briandfoy (changes requested):
Fix the die and we're good

briandfoy:
I'm going to merge this and fix up those `die`. No big whoop.

briandfoy:
I backtracked on the directories silently failing. The strings I have on FreeBSD silently ignore. If you have an example that doesn't, let me know.

PR #263, opened 2023-09-25
wc: continue if open() fails

mknos:
* Exit with failure code in usage()
* Tweak usage string: indicate zero or more files with [file...]
* Follow GNU wc: continue processing next argument if open() fails
* Delete old commented code; probably incorrect
* $opt doesn't need to be global because it is only used in the custom "Mac" code block
* Print warning for directory arguments and skip them

### DIFF
--- a/bin/wc
+++ b/bin/wc
@@ -11,29 +11,30 @@ License: perl
 
 =cut
 
-END {
-    close STDOUT || die "$0: can't close stdout: $!\n";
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
We could probably pull out all the Mac stuff since I don't think anyone would use MacPerl anymore. Not a big deal for this PR though.

PR #266, opened 2023-09-28
env: validate -u argument

mknos:
* Standard env command does not support -u NAME for unsetting a variable [1]
* OpenBSD env doesn't implement -u, but NetBSD and GNU do
* GNU env prevents argument to -u from having a '=' character
* NetBSD does the same, and its manual explicitly mentions this restriction

1. https://pubs.opengroup.org/onlinepubs/9699919799/utilities/env.html

### DIFF
--- a/bin/env
+++ b/bin/env
@@ -28,7 +28,12 @@ while ( @ARGV && $ARGV[0] =~ /^-/ ) {
 	if ( $arg eq '-i' ) {
 		%ENV = ();
 	} elsif ( $arg =~ /^-u(.*)/ ) {
-		delete $ENV{ length($1) ? $1 : shift };
+		my $val = length $1 ? $1 : shift;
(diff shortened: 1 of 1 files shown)


bin/env
```diff
@@ -28,7 +28,12 @@ while ( @ARGV && $ARGV[0] =~ /^-/ ) {
 	if ( $arg eq '-i' ) {
 		%ENV = ();
 	} elsif ( $arg =~ /^-u(.*)/ ) {
-		delete $ENV{ length($1) ? $1 : shift };
+		my $val = length $1 ? $1 : shift;
```
briandfoy, line 31: A bit ugly, but until we switch to Get::Opts, this is fine. I've done it like this before but I've never been happy with it.

briandfoy (approved):

PR #267, opened 2023-09-28
which returns directory paths

mknos:
* The following output is incorrect because the results all refer to directories

$ perl which . .. /
/usr/local/sbin/.
/usr/local/sbin/..
/usr/local/sbin//

* The command "which" is supposed to find files available to execute, not directories
* Resolve this by adding explicit directory check before checking "-x $path"
* Style: declare path variables to make code simpler to read

### DIFF
--- a/bin/which
+++ b/bin/which
@@ -100,21 +100,30 @@ foreach my $command (@ARGV) {
     }
 
     foreach my $dir (@PATH) {
+        my $path = $dir . $file_sep . $command;
+
+        if (-d $path) {
+            next;
+        }
         if ($^O eq 'MacOS') {
-            if (-e "$dir$file_sep$command") {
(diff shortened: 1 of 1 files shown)


bin/which
```diff
@@ -100,21 +100,30 @@ foreach my $command (@ARGV) {
     }
 
     foreach my $dir (@PATH) {
+        my $path = $dir . $file_sep . $command;
+
+        if (-d $path) {
+            next;
+        }
         if ($^O eq 'MacOS') {
```
briandfoy, line 108: Do you know why we'd single out MacOS here? I think this is Mac Classic, which we don't need to support. We don't need to change this for this PR, but it's something to think about. I wonder what happens on Windows, which has a completely different permissions model.

briandfoy (approved):

PR #276, opened 2023-10-02
cal: avoid split() when formatting a whole year

mknos:
* When printing month-boxes horizontally I wasn't happy with having to split each month-box string
* Make fmt_month() return a list of lines instead
* This makes it simpler to combine month-boxes line by line
* Some common code moves into print_title()

### DIFF
--- a/bin/cal
+++ b/bin/cal
@@ -41,32 +41,39 @@ if( $cal::num_args == 0 ) {
 }
 
 if( $cal::year && $cal::month ) {
-	print fmt_month($cal::year, $cal::month, 1);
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

bin/cal
```diff
@@ -41,32 +41,39 @@ if( $cal::num_args == 0 ) {
 }
 
 if( $cal::year && $cal::month ) {
-	print fmt_month($cal::year, $cal::month, 1);
+	my @lines = fmt_month($cal::year, $cal::month, 1);
```
briandfoy, line 44: Yeah, I agree with this sort of change.

PR #288, opened 2023-10-11
xargs: add -0 flag

mknos:
* The -0 option to xargs is provided by FreeBSD, NetBSD, OpenBSD and GNU
* Tested against the output of "find . -print0"
* While here add strict (code was already compliant)
* It would be good to add a POD manual here later


briandfoy (approved):
Very nice. I use -0 all the time with the real xargs.

PR #290, opened 2023-10-13
tee: null mode param passed to open()

mknos:
* Sometimes tee fails to open a file for writing because of a confusing filename guard regex

  echo yo | perl tee  . '|o|'
tee: cannot open .: Is a directory
Unknown open() mode '' at tee line 45.

* '|o|' is a valid filename on my system if it is quoted like this
* GNU tee doesn't complain about this usage (aside from complaining about the directory argument '.')
* $mode is already determined based on -a option
* Making decisions based on the filename seems incorrect so just simplify open()

### DIFF
--- a/bin/tee
+++ b/bin/tee
@@ -42,7 +42,7 @@ if ($nostdout) {
 $| = 1 if $unbuffer;
 
 for (@ARGV) {
-    if (!open($fh, (/^[^>|]/ && $mode), $_)) {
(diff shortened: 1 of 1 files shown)


mknos:
I came up with a reason why the guard code might have been useful once.
If the open() statement was written with only 2 arguments, characters like '|' in the 2nd param indicate a pipe.
There is no need for this paranoia with 3-argument open().

PR #301, opened 2023-10-22
ls: retire invalid ioctl()

mknos:
* The ioctl number assumed for TIOCGWINSZ is non-portable and would only work by accident
* The ioctl() call fails on my system because it is invalid
* Perl Cookbook example has the exact hex constant and mentions it is not portable [1]
* It seems better to remove this construct until it can be replaced with something that works

1. https://www.oreilly.com/library/view/perl-cookbook/1565922433/ch04s19.html

### DIFF
--- a/bin/ls
+++ b/bin/ls
@@ -41,6 +41,17 @@ my @perms = qw(--- --x -w- -wx r-- r-x rw- rwx);
 my @ftype = qw(. p c ? d ? b ? - ? l ? s ? ? ?);
 $ftype[0] = '';
 
+sub get_columns {
+  eval {
+    require IO::Tty;
+  } or do {
+    return 80;
+  };
+  my @winsize = IO::Tty::get_winsize(*STDIN);
+  shift @winsize; # lines
+  return shift @winsize; # columns
+}
+
 sub format_mode {
   my $mode = shift;
   my %opts = @_;
@@ -81,7 +92,7 @@ my %Options = ();	# option/flag arguments
 my $SixMonths =		# long listing time if < 6 months, else year
  60*60*24*(365/2);
 my $VERSION = '0.70';	# because we're V7-compatible :)
-my $WinCols = 80;	# window columns of output
+my $WinCols;		# window columns of output
 
 # ------ compensate for lack of getpwuid/getgrgid on some platforms
 eval { my $dummy = ""; $dummy = (getpwuid(0))[0] };
@@ -450,6 +461,7 @@ if ($Options{'f'}) {
 	$Options{'a'} = 1;
 }
 
+$WinCols = get_columns();
 $Attributes = stat(STDOUT);
 if ($Attributes->mode & 0140000) {
 	$Options{'1'} = '1';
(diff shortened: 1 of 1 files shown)


briandfoy:
Yeah, we should fix this. This patch is better than what is already happening.

However, is hard-coding 80 the way to go? Is there another way to get the terminal width in windows that works across different console applications? Maybe https://stackoverflow.com/questions/6812224/getting-terminal-size-in-c-for-windows is useful? Is that in Win32:: somewhere? stty maybe? Use Terminal::ReadKey if it is around?

If you don't want to deal with all that in this patch, that is fine, but we can open an issue for it otherwise.

bin/ls
```diff
@@ -81,14 +81,7 @@ my %Options = ();	# option/flag arguments
...
 my $VERSION = '0.70';	# because we're V7-compatible :)
-my $WinSize = "\0" x 8;	# window size buffer
-my $TIOCGWINSZ =	# get window size via ioctl()
- 0x40087468;		# should be require sys/ioctl.pl,
-			# but that won't exist on all platforms
-my $WinCols = 0;	# window columns of output
-my $WinRows = 0;	# window rows of output
-my $Xpixel = 0;		# window start X
-my $Ypixel = 0;		# window start Y
+my $WinCols = 80;	# window columns of output
```
briandfoy, line 84: Is there a way we can query the terminal and still  use this as a fallback?

briandfoy (changes requested):

briandfoy:
I marked this as "changes requested", but if you don't want to add another solution with this patch that's fine. We can definitely remove the broken method. I'd prefer another solution, but baby steps is fine. If you don't want to do that just yet, just say so and I'll merge this.

mknos:
I added a new commit. According to the documentation, CPAN module IO::Tty has been tested on many different unix systems. get_winsize() returns a sane result on my Linux system. I suspect it doesn't work out of the box on Windows; however, the old ioctl() code didn't work there either. Possibly we can add some windows-specific code in the get_columns() wrapper later.

bin/ls
```diff
@@ -41,6 +41,17 @@ my @perms = qw(--- --x -w- -wx r-- r-x rw- rwx);
 my @ftype = qw(. p c ? d ? b ? - ? l ? s ? ? ?);
 $ftype[0] = '';
 
+sub get_columns {
+  eval {
+    require IO::Tty;
```
briandfoy, line 46: IO::Tty doesn't work on Windows, and on linux and friends I think it's too much. Here are some things that will get the columns:

    stty size  | cut -d' ' -f 2

    tput cols

On Windows with powershell (with Perl post processing):

    powershell -command "&{(get-host).ui.rawui.WindowSize;}"

briandfoy (changes requested):

briandfoy:
I'll merge this and fix it up later.

briandfoy:
This is fixed by 7ee2d98a20aff9a4b644c8ff5eb455421e7dacb6.

PR #305, opened 2023-10-27
Prefer shar over par

mknos:
* par is a variant of shar which has the option -z to behave like shar
* par is perl-specific and not a standard unix command
* FreeBSD has a par command related to text formatting, but this is not common to other BSDs or Linux
* par seems a poor fit for this repo, especially given that shar works again


briandfoy:
Let me think about this for a bit. I understand your point but I don't know if I agree with it just yet. We've distributed par for so long that I might want to keep it. Going forward, I'd certainly prefer to not add special tools like that.

briandfoy:
We'll keep par, mostly for historical reasons. If someone was using it, I don't want to take it away from them. This is a good suggestion otherwise.

PR #337, opened 2023-11-21
tar: raise error for directory argument to -f

mknos:
* Adding die() for directory argument ensures tar doesn't exit with a success code
* Test command: perl tar -tf .
* Found when testing against GNU tar


briandfoy (approved):
This can be improved with the more sophisticated exits we've been doing, but this is good enough for now.

PR #348, opened 2023-11-23
fix exit status

mauke:
`exit` has higher precedence than `?:` and `==`, so `exit $Errors == 0 ? 0 : 1` parses as `(exit $Errors) == 0 ? 0 : 1`, which is not what was intended.

### DIFF
--- a/bin/install
+++ b/bin/install
@@ -82,7 +82,7 @@ else {
     install_files();
 }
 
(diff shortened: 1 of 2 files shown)


mknos:
I endorse this patch, but does the same thing need to be applied to bin/patch?

mauke:
Yes, good point.

briandfoy (approved):

PR #349, opened 2023-11-25
ed: error string consolidation

mknos:
* Reduce code noise by declaring error strings in constants
* Check close() return value in edWrite() as done in edEdit()
* Move some code to end of edEdit() that was being run for both $InsertMode and !$InsertMode branches
* Reduce layers of nesting in edQuit()

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -74,6 +74,18 @@ use strict;
 
 use Getopt::Std qw(getopts);
 
+use constant E_ADDREXT => 'too many addresses';
+use constant E_ADDRBAD => 'invalid address';
+use constant E_ARGEXT  => 'extra arguments detected';
+use constant E_SUFFBAD => 'invalid command suffix';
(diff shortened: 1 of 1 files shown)


bin/ed
```diff
@@ -74,6 +74,18 @@ use strict;
...
+use constant E_ARGEXT  => 'extra arguments detected';
+use constant E_SUFFBAD => 'invalid command suffix';
+use constant E_CLOSE   => 'cannot close file';
+use constant E_OPEN    => 'cannot open file';
+use constant E_READ    => 'cannot read file';
+use constant E_NOFILE  => 'no current filename';
+use constant E_UNSAVED => 'buffer modified';
+use constant E_CMDBAD  => 'unknown command';
+use constant E_PATTERN => 'invalid pattern delimiter';
+
```
briandfoy, line 88: Nice change!

briandfoy (approved):

PR #354, opened 2023-11-30
ed: simplify code for r command

mknos:
* For "0r filename", file content is added to the beginning of lines list (preserving initial undef element)
* For "3r filename", file content is added after the 3rd line in buffer (i.e. after element 4 considering initial undef element)
* Each case can be written as a single call to splice (offset is 0+1 for "0r" case  and 3+1 for "3r" case)


briandfoy (approved):
So much can be saved if people where comfortable with slice()

PR #365, opened 2023-12-04
grep: fix decompression feature

mknos:
* By default, this version of grep detects compressed files and decompresses them by running a command via a pipe
* This feature is not standard for grep, and possibly in future we could add an option to disable it
* I suspect this was broken by previously converting the script to 3-argument open() --- restore 2-argument open() only for this decompression case
```
%gzip -k bc
%perl grep sub bc.gz
grep: zcat < bc.gz |: No such file or directory
```

### DIFF
--- a/bin/grep
+++ b/bin/grep
@@ -274,6 +274,7 @@ sub matchfile {
     $total = 0;
 
 FILE: while (defined ($file = shift(@_))) {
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
I agree that we should remove that feature.

PR #375, opened 2023-12-13
find: simpler error message

mknos:
* It is good enough for awk front-end to print a short error with the name of the required CPAN module
* Make the error message printed by find front-end consistent with this

%perl find 
This program needs the App::find2perl module.
%perl awk
This program needs the App::a2p module.


briandfoy (approved):
Yeah, much more user-friendly. Thanks,

PR #376, opened 2023-12-14
apply: capture system() failure

mknos:
* According with OpenBSD, "apply 1 2 3" should print an error message each time "1" command fails to run, then exit with error code
* "apply false 2 3" should not print an error message because "false" runs successfully, then exit with error code
* The return value from system() was not checked previously so no feedback was provided if something went wrong
* Prefix "usage:" to the usage string
* Bump version
* In future it would be useful to add debug option (-d) to show but not run the target commands

### DIFF
--- a/bin/apply
+++ b/bin/apply
@@ -14,20 +14,13 @@ License: perl
 
 use strict;
 
+use File::Basename qw(basename);
+
 use constant EX_SUCCESS => 0;
 use constant EX_FAILURE => 1;
 
-my ($VERSION) = '1.2';
-
-sub usage () {
-    require File::Basename;
-    $0 = File::Basename::basename ($0);
-    print <<EOF;
-$0 (Perl bin utils) $VERSION
-$0 [-ac] [-#] command argument [argument ...]
-EOF
-    exit EX_FAILURE;
-};
+my $Program = basename($0);
+my ($VERSION) = '1.3';
 
 my $argc  = 1;
 my $magic = '%';
@@ -48,7 +41,7 @@ while (@ARGV) {
         next;
     }
     if ($ARGV [0] =~ /^-/) {
-        usage;  # usage will exit;
+        usage();  # usage will exit
     }
     last;
 }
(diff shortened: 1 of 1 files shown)


bin/apply
```diff
@@ -65,23 +58,43 @@ if (@thingies) { # Ignore $argc, found our own.
...
             shift;
         }
     }
 }
-exit EX_SUCCESS;
+exit $err;
+
+sub checkerr {
+    my $status = shift;
+    if ($status != 0) {
```
briandfoy, line 85: This is a little weird, because not every command returns 0 for success. Some commands have more than one success value. 

If this is the way actual `apply` works though, we can just copy that even if it misses edge cases.

briandfoy (changes requested):

bin/apply
mknos: In the OpenBSD code, running the command is handled in the function mysystem() [1].
If fork() fails, mysystem() returns -1.
If fork() succeeds and execl() fails in the child process, the child will exit with status 1 (fetched via waitpid).
If execl() succeeds in the child process, but then waitpid() fails, pid is updated and mysystem() returns -1.
Otherwise, mysystem() returns the value from waitpid().

Return value of 0 from mysystem() is treated as "success".
Since it is called in a loop, if any instance of mysystem() returns non-zero, apply will do exit(1).
It doesn't matter if the executed command used exit code 1 or 100, apply will exit with 1.
The FreeBSD, NetBSD and OpenBSD manuals don't appear to mention this.

In the perl code, if system() returned -1 then likely either fork() or waitpid() failed.
Printing an error here is close to the C version because the command could not be started for some reason (e.g. not found).
An error is not printed if the command had exit code 100, because this information is passed on *lossily* by the eventual exit code of 1.

1. https://cvsweb.openbsd.org/cgi-bin/cvsweb/src/usr.bin/apply/apply.c?annotate=1.29

briandfoy: Thanks for the code pointers. I think that code is being less thoughtful than it could be, but that doesn't matter because we are copying what it does and not what we think it should do.

PR #388, opened 2023-12-25
cmp: avoid repeated substr() on binary buffer

mknos:
* Slightly simplify code by unpack()ing buffers that are not equal
* $lines_read becomes a tally, and $report_lines is always ahead by one because $lines_read counts from zero
* When printing the differences with -l flag, ord+substr pattern is avoided by using unpacked buffers
* Retire EX_USAGE which had same value as EX_FAILURE


briandfoy:
Let's keep EX_USAGE. It doesn't matter that it has the same value as something else. It's possible to have a different value, is it's existence is a logical win. I'll fix up the patch because the rest is good.

briandfoy (approved):

PR #392, opened 2023-12-29
fortune: fail for bad options

mknos:
* Print usage and exit if an unknown option was entered, following OpenBSD version[1]
* -h is not really an option
* Mention -n option in usage string as done in SYNOPSIS
* Document the purpose of -d flag in usage string and pod

1. https://github.com/openbsd/src/blob/master/games/fortune/fortune/fortune.c#L304

### DIFF
--- a/bin/fortune
+++ b/bin/fortune
@@ -42,7 +42,7 @@ my $STR_ROTATED = 0x4;
 # Globals
 my (%opts);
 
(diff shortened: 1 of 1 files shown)


briandfoy:
Wow, fortune has an enhanced debugging mode. Never knew that.

briandfoy (approved):

PR #398, opened 2024-01-03
Use two-argument form of FileHandle->new

mknos:
* Previously the calls to open() were standardised to have 3 arguments
* Some scripts open files via FileHandle->new() instead of open()
* The pr command already used the preferred 2-argument form; also switch ar and file commands over to this


briandfoy (approved):
Good catch

PR #399, opened 2024-01-03
tail: allow -n1 and -n 1

mknos:
* I habitually type "tail -n5"; this is supported by GNU tail and OpenBSD tail
* Enabling bundling option in Getopt::Long makes it work in this version
* Style: avoid capture variable when the entire argument string "+590" is passed to check_number()
* test1: "perl tail -5 tail" -- historical option format, last 5 lines
* test2: "perl tail +590 tail" -- historical option format, line 590 to end
* test3: "perl tail -n5 tail" -- last 5 lines
* test4: "perl tail -n-5 tail" -- last 5 lines
* test5: "perl tail -n+590 tail" -- line 590 to end

### DIFF
--- a/bin/tail
+++ b/bin/tail
@@ -82,8 +82,8 @@ sub parse_args()
 
     } else {
 
-    # special handling for -numeric options
-        Getopt::Long::config("pass_through");
+    # pass_through is special handling for -numeric options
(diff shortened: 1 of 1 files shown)


briandfoy:
Does this change need to happen in head too?

PR #403, opened 2024-01-09
cut: better input validation for -b and -f

mknos:
* Options -b and -f are consistent in how the value is handled
* Byte and field numbers count from 1
* Negative numbers are treated as a range end, with range starting from 1
* Regular ranges have 2 numbers separated by '-'
* Introduce a number validation function
* test1: positive -b value: "echo 'a:b:c:d' | perl cut -b 3" --> byte 3 is "b"
* test2: negative -b value: "echo 'a:b:c:d' | perl cut -b -3" --> bytes 1-3 are "a:b"
* test3: -b range: "echo 'a:b:c:d' | perl cut -b 2-3" --> bytes 2-3 are ":b"
* test4: positive -f value: "echo 'a:b:c:d' | perl cut -d ':' -f 2" --> field 2 is "b"
* test5: negative -f value: "echo 'a:b:c:d' | perl cut -d ':' -f -2" --> fields 1-2 are "a:b"
* test6: -f range: "echo 'a:b:c:d' | perl cut -d ':' -f 2-3" --> fields 2-3 are "b:c"

### DIFF
--- a/bin/cut
+++ b/bin/cut
@@ -21,7 +21,6 @@ License: perl
 #   Perl Power Tools -- http://language.perl.com/ppt/
 #
 
-$^W = 1;    # -w
 use strict;
 
 use File::Basename qw(basename);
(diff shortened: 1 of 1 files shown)


briandfoy (approved):
I think there are a few formatting things to fix up, but I'll take care of that. Maybe it's just the GitHub display of the diff.

PR #405, opened 2024-01-10
ed: implement ! command

mknos:
* Take a hint from the TODO comments
* The big regex in edParse() doesn't help because spaces have a special meaning; the new branch bypasses it
* "!" line is printed at end of command output (or after error message if command can't be started)
* Ignore if the command exits with non-zero status; print error only if command doesn't start
* edPipe() doesn't take any line addresses because the output is not saved into the buffer
* Tested against GNU and OpenBSD versions
* test1: shell pattern: "!for i in 1 2 3; do echo $i; done"
* test2: zero exit status command: "!ls"
* test3: non-zero exit status command: "!false"

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -60,7 +60,6 @@ License: gpl
 #                u - undo
 #                v - global command "inVerted"
 #                x - get encryption key
-#                ! - shell command
 #
 #        - Create regression test suite...test against "real" ed.
 #        - add a "-e" flag to allow it to be used in sed(1) like fashion.
@@ -230,6 +229,8 @@ while (1) {
 
         if (!length($command)) {
             &edSetCurrentLine;
+        } elsif ($command eq '!') {
+            edPipe();
         } elsif ($command eq '=') {
             &edPrintLineNum;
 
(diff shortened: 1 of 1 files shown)


bin/ed
```diff
@@ -395,6 +396,15 @@ sub escape_line {
     return $s;
 }
 
+# does not modify buffer
+sub edPipe {
+    return unless defined $args[0];
+    my $rc = system $args[0];
```
briandfoy, line 402: Let's be a lot more careful here. This is `system` in the single argument form, which will have the shell translate the string before running it. Let's either check the string for safety, or convert this to the list form with the braces syntax.

briandfoy (changes requested):

bin/ed
mknos: > test1: shell pattern: "!for i in 1 2 3; do echo $i; done"
In my understanding, the "!" sub-command in ed is expected to run a shell command, or the above example wouldn't work. 
This works the same in both OpenBSD ed and GNU ed. In terms of being careful, GNU has a restricted mode which is enabled by running "red" or "ed -r". The "!" sub- command is not available in restricted mode. OpenBSD doesn't have this mode; probably because a text editor doesn't need to be ultra secure, and no one will want to run restricted-ed anyway.

briandfoy: After thinking about this, I agree. It's what the command does. Thanks for looking at it again. Maybe we can add it later. I've added #409 for this.

PR #412, opened 2024-01-19
sort: enforce minimum value of 1 for -F and -y

mknos:
* Basic validation
* Options -F and -y are non-standard and control allocation limits within sort
* Currently if the value of zero is given, it is ignored because of truth-test
* Negative amounts are accepted but don't make sense in the context of a limit size

### DIFF
--- a/bin/sort
+++ b/bin/sort
@@ -85,9 +85,14 @@ sub _sort_file {
 
     # set output and other defaults
     $opts->{o}   = !$opts->{o} ? '' : $opts->{o};
-    $opts->{'y'} ||= $ENV{MAX_SORT_RECORDS} || 200000;  # default max records
-    $opts->{F}   ||= $ENV{MAX_SORT_FILES}   || 40;      # default max files
-
+    if (defined $opts->{'F'}) {
+        die "option -F expects a positive number\n" if (int($opts->{'F'}) < 1);
+    }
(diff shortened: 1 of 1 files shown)


bin/sort
```diff
@@ -85,9 +85,14 @@ sub _sort_file {
...
-    $opts->{F}   ||= $ENV{MAX_SORT_FILES}   || 40;      # default max files
-
+    if (defined $opts->{'F'}) {
+        die "option -F expects a positive number\n" if (int($opts->{'F'}) < 1);
+    }
+    if (defined $opts->{'y'}) {
+        die "option -y expects a positive number\n" if (int($opts->{'y'}) < 1);
+    }
+    $opts->{'y'} ||= $ENV{'MAX_SORT_RECORDS'} || 200000;  # default max records
+    $opts->{'F'} ||= $ENV{'MAX_SORT_FILES'}   || 40;      # default max files
```
mknos, line 95: If this was structured differently the validation could also cover the ENV input. Should I add another commit?

briandfoy: I'd rewrite all that to set defaults in `$opts` first, then process args, then check args. You'd still want to check the value of `y` after you let someone set it through an environment variable.

briandfoy (changes requested):

briandfoy (approved):

PR #413, opened 2024-01-23
ed: simplify relative line calculation

mknos:
* Based on prior regex, "+-" and "-+" are invalid relative addresses; GNU ed accepts "+-" but this version treats it as "extra arguments" error
* "++n" is numbered print, 2 lines ahead of currentLine
* "--n" is numbered print, 2 lines behind currentLine
* Only one of $pluses or $minuses will be set at this time
* Writing the logic in this way will make it easier to support "--++" in future for GNU compat


briandfoy (approved):
Yeah, that's much easier to figure out.

PR #416, opened 2024-01-25
hangman versus small file

mknos:
* When testing hangman with a small word list the game would terminate before it even started 
* get_a_word() assumed that the wordlist file had 9151 words in it
* This would be a problem for larger word lists too; the list is effectively truncated
* Scale random index ($val) based on the number of lines in word list (with penalty of reading list twice)
* Filehandle FILE used in close() was incorrect
* Raise error if close() failed
* Test1: wordlist.txt with only 2 words
* Test2: wordlist.txt from dict file on my linux system (101,924 words)

### DIFF
--- a/bin/hangman
+++ b/bin/hangman
@@ -72,11 +72,18 @@ print "\nTHANKS FOR PLAYING!\n";
 
 sub get_a_word {
 	my $wordlist = "wordlist.txt";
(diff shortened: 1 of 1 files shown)


briandfoy (approved):
That's such an oddly specific number.

PR #417, opened 2024-01-25
echo: remove non-standard -?

mknos:
* Sometimes echo is built into the shell
* Standards document for echo doesn't mention the -? flag [1]
* The versions of echo that I tested don't support it
* I think it's better to follow convention and not treat -? as an option
* Sync pod and fix a typo in DESCRIPTION

1. https://pubs.opengroup.org/onlinepubs/009604599/utilities/echo.html

```
# bash builtin
% builtin echo $BASH_VERSION
5.1.4(1)-release
% builtin echo -?
-?
% builtin echo --version
--version

# gnu echo
% /bin/echo -?
-?
% /bin/echo --version
echo (GNU coreutils) 8.32

# pdksh builtin
% builtin echo -?
-?
% builtin echo --version
--version
% builtin echo $KSH_VERSION
@(#)PD KSH v5.2.14 99/07/13.2

# openbsd echo
% /bin/echo -?
-?
% /bin/echo --version
--version
```

### DIFF
--- a/bin/echo
+++ b/bin/echo
@@ -15,33 +15,15 @@ package PerlPowerTools::echo;
 
 use strict;
 
-my ($VERSION) = '1.2';
+my ($VERSION) = '1.3';
 
 __PACKAGE__->run( @ARGV ) unless caller();
 
 sub run {
     my ( $self, @args ) = @_;
 
-    unless (@args) {
-        print "\n";
-        exit 0;
-    }
-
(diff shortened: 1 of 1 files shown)


briandfoy (approved):
In general, I think your changes to let `-?` be handled as an unknown option and trigger `usage` is a good way to go.

PR #418, opened 2024-01-27
ed: eliminate global variable $fname

mknos:
* edEdit() is always called and expects a filename to be passed in $args[0]
* Special filename '-' is the same as no filename; edEdit() expects $args[0] to be unset and will read stdin
* Extra file arguments are not allowed; only one file can be edited at a time
* When stdin is read, byte counts are not displayed

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -163,8 +163,8 @@ if ($opt{'v'}) {
     $EXTENDED_MESSAGES = 1;
 }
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy (commented):
Approved. Why are there no options on this dialog?

briandfoy:
changes: internal improvements

PR #419, opened 2024-01-27
head: support -NUM

mknos:
* The POD text lied because the code was not compatible with OpenBSD
* Make head consistent with tail by accepting the historic -NUM form as -n NUM
* While here, correct the name of helper function: s/tail_fh/head_fh/

```
%perl head -2 head
#!/usr/bin/perl

%perl head -n 2 head
#!/usr/bin/perl

%perl tail -2 tail
=cut

%perl tail -n 2 tail
=cut

```

### DIFF
--- a/bin/head
+++ b/bin/head
@@ -23,8 +23,7 @@ use constant EX_FAILURE => 1;
 my $Program = basename($0);
 my ($VERSION) = '1.3';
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):
heh, `tail_fh`. It's okay to have separate pull requests, but I think many of the changes for these two programs have been the same and you could combine them in one PR. Either way is fine with me.

briandfoy:
changes: handle - option to be like -n <DIGIT>

PR #420, opened 2024-01-27
tail: bad option combo

mknos:
* Handle historic -NUM option before calling GetOptions(); -5 is -n5
* The options -b, -c and -n are incompatible; print usage and exit as done in OpenBSD tail
* Clarify an error message: s/null/zero/
```
# valid
perl tail -n2 tail | wc -c # last lines
6
perl tail -c2 tail | wc -c # last bytes
2
perl tail -b2 tail | wc -c # last blocks (512 bytes)
1024

# invalid
perl tail -b2 -c2
perl tail -b2 -n2
perl tail -c2 -n2 
perl tail -b2 -2
```

### DIFF
--- a/bin/tail
+++ b/bin/tail
@@ -82,6 +82,8 @@ sub parse_args()
 
     } else {
 
+    @ARGV = map { m/\A([\-\+][0-9]+)\Z/ ? "-n$1" : $_ } @ARGV; # historic
+
     # pass_through is special handling for -numeric options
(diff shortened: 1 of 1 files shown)


briandfoy (approved):
heh, for some reason `new_argv` made me laugh :)

briandfoy:
changes: handle -<DIGIT> option to be like `-n <DIGIT>`

PR #421, opened 2024-01-28
rm: document -v

mknos:
* Add pod description for -v flag and include it in usage string 
* Currently nothing calls usage(), but this is still helpful for reading pod2text

### DIFF
--- a/bin/rm
+++ b/bin/rm
@@ -19,7 +19,7 @@ rm - remove directory entries
 
 =head1 SYNOPSIS
 
-	rm [-fiPrR] file ...
+	rm [-fiPrRv] file ...
 
 =head1 DESCRIPTION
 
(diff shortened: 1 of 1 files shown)


bin/rm
```diff
@@ -283,7 +283,7 @@ sub usage {
 sub error_fh { my $self = shift; $self->{error_fh} }
 sub error {
 	my $self = shift;
-	print { $self->error_fh || * STDERR } @_;
+	print { $self->error_fh || *STDERR } @_;
```
briandfoy, line 286: This is fine for now, but I'd move that default value into `error_fh`.

briandfoy (approved):

briandfoy:
changes: document the -v switch

PR #423, opened 2024-01-28
cat: explicitly close files

mknos:
* Guarantee that if close() failed for file1 that we print a warning for file1 before proceeding to process file2, then eventually exit with error code
* To preserve "cat - -", make sure *STDIN is not closed (on my linux system I type ctrl-d twice to terminate do_file() with EOF)

### DIFF
--- a/bin/cat
+++ b/bin/cat
@@ -93,6 +93,10 @@ sub do_file {
 
         print;
     }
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: explicitly close files

PR #424, opened 2024-01-28
install: "usage()" > "die usage()"

mknos:
* Making usage() exit instead of returning a string is more consistent with usage() in other commands
* test1: "perl install -G" --> unknown option
* test2: "perl install" --> no arguments
* test3: "perl -c -d ABC DEF" --> -d not allowed with -c
* test4: "perl install awk xargs" --> final argument is not a directory


briandfoy (approved):

briandfoy:
changes: some internal improvements on usage handling

PR #425, opened 2024-01-28
fold: terminate options with --

mknos:
* Tested against GNU fold and OpenBSD fold
* POSIX semantics: If I want to fold a file called "-s" I put it after "--" (perl fold -s -- -s)
* Single "-" means read stdin; "---" is an invalid option
* Based on a similar patch to uniq


briandfoy (approved):

briandfoy:
changes: terminate command-line options with --

PR #427, opened 2024-01-29
wc: terminate options with --

mknos:
* Not all commands know about "--" yet
* Conversion to Getopt::Std is simple for wc because options are all single letter flags
* Remove differences for MacPerl which likely don't matter anymore (Getopt::Std will work on OSX perl)
* test1: "perl wc -c -- -c" --> process a file called "-c"
* test2: compare output of GNU version (below); different count for -w flag was present in previous versions
```
%for opt in c l m w; do echo "GNU -$opt"; wc -${opt} awk; echo "PPT -$opt"; perl wc -${opt} awk; echo; done GNU -c
4595 awk
PPT -c
      4595 awk

GNU -l
204 awk
PPT -l
       204 awk

GNU -m
4595 awk
PPT -m
       204       675      4595      4595 awk

GNU -w
701 awk
PPT -w
       675 awk
```

### DIFF
--- a/bin/wc
+++ b/bin/wc
@@ -15,47 +15,20 @@ use strict;
 use locale;
 
 use File::Basename qw(basename);
+use Getopt::Std qw(getopts);
 
 use constant EX_SUCCESS => 0;
 use constant EX_FAILURE => 1;
 
 my $Program = basename($0);
 
-my %opt;
-
 sub usage {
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: terminate argument processing with --

PR #428, opened 2024-01-29
grep: GREP_OPTIONS preferred env name

mknos:
* NetBSD and FreeBSD versions of grep have an environment variable called GREP_OPTIONS which is added before any options given on the commandline
* This version did the same thing for env variable TCGREP so it could be renamed for slightly better BSD compatibility
* Update pod to explain how commandline options are used with respect to GREP_OPTIONS
* Referring to grep as tcgrep in the pod manual is confusing because the program is installed as grep not tcgrep; users might looks for tcgrep and not find it

### DIFF
--- a/bin/grep
+++ b/bin/grep
@@ -117,7 +117,7 @@ Specials:
 	q   quiet about failed file and dir opens
 	T   trace files as opened
 
-May use a TCGREP environment variable to set default options.
+May use GREP_OPTIONS environment variable to set default options.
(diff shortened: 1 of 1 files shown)


briandfoy (approved):
I guess tcgrep is "Tom Christiansen grep"

briandfoy:
changes: remove TCGREP environment var, use GREP_OPTIONS instead

PR #429, opened 2024-01-30
id: retire non-standard -h

mknos:
* With this patch, usage string is still printed for -h, but explicit code is not needed for it
* Standards document does not mention -h [1]
* Fix a typo in the pod document

1. https://pubs.opengroup.org/onlinepubs/009696899/utilities/id.html

### DIFF
--- a/bin/id
+++ b/bin/id
@@ -21,10 +21,9 @@ License:
 
 use strict;
 use Getopt::Std;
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: remove -h option, although you still get the usage statement

PR #430, opened 2024-01-30
sleep: fail for extra args

mknos:
* Align sleep with OpenBSD version, which shows usage string if extra arguments are given
* Previously this version ignored extra arguments
* Make pod text more useful
* test1: "perl sleep" --> no argument
* test2: "perl sleep -x" --> bad option
* test3: "perl sleep -- 1" --> valid: supported by GNU and OpenBSD
* test4: "perl sleep -- -1" --> bad interval
* test5: "perl sleep -- 1 2" --> too many arguments

### DIFF
--- a/bin/sleep
+++ b/bin/sleep
@@ -13,31 +13,36 @@ License: perl
 
 use strict;
 
-my ($VERSION) = '1.202';
+use File::Basename qw(basename);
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: fail for extra arguments

PR #431, opened 2024-01-30
fmt: improve -NUM option parsing

mknos:
* This version of fmt uses Getopt::Long to validate that the -w argument is a number
* Problem1: -NUM in the wrong place is treated as a valid option
*  e.g. with "perl fmt A -1 B", the expected input files are A, '-1' and B (-1 is not an option because A terminates the options)
* Problem2: -w argument needs a space (GNU fmt allows -w40)
* Now "-10", "-w10" and "-w 10" are treated the same

### DIFF
--- a/bin/fmt
+++ b/bin/fmt
@@ -25,18 +25,10 @@ my $Program = basename($0);
 
 my $MAX_WIDTH = 75;
 
-# Take care of special case, bare -width option:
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: handle -<DIGIT> as option instead of a file

PR #432, opened 2024-01-31
ed: simplify debug

mknos:
* Simplify -d option to become a flag and not take an argument
* The only supported argument was "parse", so this patch makes debugging easier to enable
* The old way with -ddebugstring did not provide any feedback if I typed an unsupported string
* Sync usage string in Usage() and Synopsis
* Remove redundant comment where usage string was also being maintained (but wasn't up to date with -s or -p)

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -29,15 +29,11 @@ License: gpl
 #        - Because working on this is more fun than Y2K testing !
 #
 # Who:
-#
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: simply the debug facility

ronsavage:
Hi brian

Should the last commit read '... simplify ...'?

PR #433, opened 2024-01-31
od: refactor default output mode

mknos:
* Dedicated if-statement is not required for setting -o as the default output format for od; use the default case for setting $fmt instead


briandfoy (approved):

briandfoy:
changes: some internal improvements

PR #434, opened 2024-01-31
od: incorrect file offsets after first 16 bytes

mknos:
* The file offsets incorrectly jumped from 0 to 17, instead of 0 to 16
* When debugging this I found the formatting routines were modifying the value of $data by appending pad bytes
* Fix offsets issue by not modifying the value of $data
* Using global $len in the format routines is also not correct since the input loop was previously changed and $len will always be 1 (padding was added incorrectly in some cases)
* Remove global variable $upformat and provide the unpack format directly to unpack()
* Remove unsupported -a flag from the SYNOPSIS; it was not mentioned in help()


briandfoy (approved):

briandfoy:
This fails the tests for od since the output has changed, but I'll fix those.

briandfoy:
changes: fix offset numbering

PR #437, opened 2024-02-02
ed: clean some redundant code

mknos:
* In edMove(), special code isn't needed for $count of one because the .. operator can handle 1..1
* edPrint() already relies on this because it loops with "$adrs[0] .. $adrs[1]", where both can be equal
* Tested with command "1m20" to move the 1st line in buffer ($end is inferred from $start)
* Remove chmod() comment in edWrite(); OpenBSD version does not perform chmod() and that is a better reference
* Standards document mentions nothing about changing file modes/permissions[1]
* In edHelp() the saved error will be printed each time the "H" command enables the help flag; avoid extra return statement

1. https://pubs.opengroup.org/onlinepubs/9699919799/utilities/ed.html

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -324,9 +324,8 @@ sub edHelp {
     }
     if ($toggle) {
         $EXTENDED_MESSAGES ^= 1;
-        return unless $EXTENDED_MESSAGES;
     }
-    if (defined $Error) {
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: clean up some internals

PR #438, opened 2024-02-02
arch: forbid command arguments

mknos:
* Make arch command more compatible: no arguments are allowed
* Incorrect options were ignored but should raise an error
* GNU and OpenBSD versions raise an error for these 2 cases; OpenBSD prints usage string
* test1: "perl arch -k arch" --> valid option, invalid argument
* test2: "perl arch -x" --> invalid option

### DIFF
--- a/bin/arch
+++ b/bin/arch
@@ -12,7 +12,19 @@ License:
 =cut
 
 use strict;
+
+use File::Basename qw(basename);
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: reject all arguments

PR #439, opened 2024-02-05
rev: terminate options with --

mknos:
* GNU and OpenBSD versions of rev use getopt(), so they support '--' for terminating options; add getopt() here too
* Remove --help (-h); usage string is still printed
* Update SYNOPSIS and usage() to indicate multiple input files can be processed
* test1: "perl rev -- -v 01 02" --> 3 input files (now I can read file called "-v" without typing "./-v")
* test2: "perl rev -x" --> bad option

### DIFF
--- a/bin/rev
+++ b/bin/rev
@@ -14,6 +14,7 @@ License: gpl
 use strict;
 
 use File::Basename qw(basename);
+use Getopt::Std qw(getopts);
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: terminate args with --

PR #440, opened 2024-02-06
paste: reject implied file argument

mknos:
* Behave more like standard paste; zero arguments is considered invalid usage [1]
* Update SYNOPSIS and usage string to indicate a file argument is needed
* Add description in pod to explain special '-' file argument
* Style: Merge declaration+assignment of file handle in loops
* test1: "perl paste" --> now print usage
* test2: "perl paste paste" --> same as cat
* test3: "cat paste | perl paste -" --> same as test2
* test4: "ls | perl paste - - -" --> ls output 3 items at a time

1. https://pubs.opengroup.org/onlinepubs/7908799/xcu/paste.html

### DIFF
--- a/bin/paste
+++ b/bin/paste
@@ -24,19 +24,20 @@ my $Program = basename($0);
 my ($VERSION) = '1.3';
 my (@fh, @sep, %opt);
 
-unless (getopts('d:s', \%opt)) {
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: require at least one filename; - is standard input

PR #441, opened 2024-02-07
ed: 1s/pat1/pat2/ incorrectly processes multiple lines

mknos:
* If current line is 10, and I enter 1s/e/E/, the substitution was occurring for lines 1-10 instead of just line 1
* If I want substitution on lines 1-10 I'm supposed to type 1,10s/e/E/
* If I only want to substitute on the current line I omit the addresses and type s/e/E/
* Patch is based on existing behaviour of edPrint()
* When searching for assignments of $adrs[1] I didn't find any more instances where it's set to current line
* Add validation to prevent line address 0 for "s" command (0 is valid for a few ed commands though)
* Fix logic in edHelp() for "h" command

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -501,7 +501,7 @@ sub edSubstitute {
     $adrs[0] = $CurrentLineNum unless (defined($adrs[0]));
     $adrs[1] = $adrs[0] unless (defined $adrs[1]);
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: fix applying a substitution to a single line

PR #442, opened 2024-02-08
cat: line number compat

mknos:
* When printing line number prefix for -n and -b flags, terminate the line number with a tab character
* This makes the output compatible with the OpenBSD and GNU versions
* Flags -n and -b are not covered in standards document [1]
* Line number can be printed directly without modifying the value of line buffer $_
* Now md5sum matches for /bin/cat -n and -b output on my linux system

1. https://pubs.opengroup.org/onlinepubs/9699919799/utilities/cat.html

### DIFF
--- a/bin/cat
+++ b/bin/cat
@@ -78,9 +78,6 @@ sub do_file {
             $was_empty = $is_empty;
         }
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: terminate line number with a tab

PR #443, opened 2024-02-08
maze: tweak usage string

mknos:
* Program name comes after the "usage: " prefix
* Init maze data without using a counter variable


briandfoy (approved):

briandfoy:
changes: fix the usage message

PR #444, opened 2024-02-08
bc: standard behaviour for '-' argument

mknos:
* Standards document does not mention special interpretation of file argument '-' for bc [1]
* Argument '-' is treated as a literal filename on FreeBSD, OpenBSD and Linux
* Standard input is only read if argument list is empty
* Bypass call to open() if file argument is a directory
* Update pod SYNOPSIS

### DIFF
--- a/bin/bc
+++ b/bin/bc
@@ -27,7 +27,7 @@ Run a bc program from FILE
 
 Run a bc program from standard input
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: the - is a literal filename now

PR #445, opened 2024-02-08
id: don't suggest -a in usage

mknos:
* Improve the pod SYNOPSIS by adding the different usages  which were listed in help()
* Usage string should not include -a option because it is ignored (as documented in pod)
* getopts() parameter only needs one 'a'


briandfoy (approved):

briandfoy:
changes: note that -a is ignored

PR #446, opened 2024-02-09
nl: validate -n option

mknos:
* Expand input validation code to cover -n option
* As mentioned by the usage string, only 3 values for -n are accepted
* This was found when testing against GNU nl


briandfoy (approved):

briandfoy:
changes: validate the -n argument

PR #447, opened 2024-02-10
fold: '-' is not stdin

mknos:
* GNU fold treats '-' as a special file argument but that is an extension
* Standards document contains wording that stdin is only used if no arguments are given; OpenBSD follows this [1]
* The pod manual also says stdin is only used as a default for no file arguments ('-' is not documented to have a special meaning)
* This change was recently made to bin/head also

1. https://pubs.opengroup.org/onlinepubs/009695299/utilities/fold.html

### DIFF
--- a/bin/fold
+++ b/bin/fold
@@ -90,10 +90,12 @@ if ($Space_Break && $Width < $Tabstop) {
     usage("width must be greater than $Tabstop with the -s option");
 }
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: don't treat - as standard input

PR #448, opened 2024-02-10
basename/dirname: exit(1) for no args

mknos:
* This change is covered by Exit Status section of these documents:
https://pubs.opengroup.org/onlinepubs/007904975/utilities/dirname.html
https://pubs.opengroup.org/onlinepubs/007904975/utilities/basename.html

* This was found when testing against the GNU version
* While here, prefix usage string with "usage: "


briandfoy (approved):

briandfoy:
changes: exit with an error for no arguments

PR #449, opened 2024-02-10
fish: reduce loops

mknos:
* Init the deck without a loop
* Loop in fisher_yates_shuffle() can be written without a "next" statement; swap if unequal


briandfoy (approved):

briandfoy:
changes: reduce some internal complexity

PR #450, opened 2024-02-10
cmp: check length off-by-one

mknos:
* $checklength is the maximum offset in buffer to check which byte(s) in the 2 buffers are not equal
* The subsequent loop is from 0 to $checklength, so avoid potentially accessing $bytes1[$chunk_size]
* If $read_in1 and $read_in2 are equal to $chunk_size, the value of $checklength is not decremented
* Calculate minimum of $read_in1 and $read_in2 (these will never exceed $chunk_size), then decrement by 1

### DIFF
--- a/bin/cmp
+++ b/bin/cmp
@@ -38,6 +38,7 @@ use strict;
 
 use File::Basename qw(basename);
 use Getopt::Std qw(getopts);
+use List::Util qw(min);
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: fix off-by-one error

PR #451, opened 2024-02-10
bc: unused var

mknos:
* Set up signal handlers before running any code
* Global $YYTABLESIZE is unused
* Delete commented code related to old goto-label "outer"


briandfoy (approved):

briandfoy:
changes: a couple of internal cleanups

PR #452, opened 2024-02-12
bc: array assignment incorrectly prints

mknos:
* In exec_stmt(), instruction 'p' takes an array name and index number, then pushes the array element onto the stack (implicit print)
* The default value for array elements is 0
* The instruction '=P' assignd a value from the stack into an array element
* Make the code for 'p' simpler by appending '[]' to $name, as done for '=P' instruction
* For instruction '=P' the value being assigned to array should not be printed (this is controlled by $return variable)
* For input "a[9] = 123; a[9];" the value of 123 should be printed once, for the 2nd statement
* Tested this input against GNU and OpenBSD versions

### DIFF
--- a/bin/bc
+++ b/bin/bc
@@ -2496,16 +2496,17 @@ sub exec_stmt
      }
 
 #     debug {"p: $name, $idx.\n"};
-     unless (defined($sym_table{$name.'[]'})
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: don't print value after array element assignment

PR #453, opened 2024-02-12
whoami: treat arguments as error

mknos:
* Follow OpenBSD and GNU versions of 'whoami' by raising an error if argument count is not zero
* Add symbolic exit codes, as done in other scripts


briandfoy (approved):

briandfoy:
changes: don't allow arguments

PR #454, opened 2024-02-12
kill: signal number validation

mknos:
* The existing validation for signal-number-too-big doesn't work correctly when "use integer" is enabled
* I tested this with a very large number
* "perl kill -7685876587234234 12" proceeded to kill process 12 incorrectly
* Removing integer pragma fixes this issue for me

### DIFF
--- a/bin/kill
+++ b/bin/kill
@@ -18,7 +18,6 @@ License:
 
 use strict;
 use Config;
-use integer;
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: validate signal number arguments

PR #455, opened 2024-02-12
yes: remove GNU extension

mknos:
* Align usage with Free/Net/Open/DragonFly BSDs and Mac-osx
* Command options are not supported: "yes --hey" prints "--hey"
* String printed is the 1st command argument; extra arguments are ignored: "yes a b" prints "a"

### DIFF
--- a/bin/yes
+++ b/bin/yes
@@ -13,36 +13,10 @@ License: perl
 
 use strict;
 
-my ($VERSION) = '1.2';
+my ($VERSION) = '1.3';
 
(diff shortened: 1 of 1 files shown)


briandfoy (changes requested):
Since you are completely replacing the source, you should give yourself credit in the metadata too.

briandfoy (approved):

briandfoy:
changes: remove GNU extensions (we are a BSD clone)

PR #456, opened 2024-02-12
what: do not read stdin

mknos:
* The pod manual doesn't mention reading from standard input
* The version on my OpenBSD system prints usage if no arguments are given
* Standards document describes file argument as being required (no implicit stdin) [1]
* Standards document explicitly says standard input is not used, i.e. no special argument or option to signify stdin
* When testing this I discovered invalid option -x did not raise an error, so add error handling
* Below output is from my linux system: two of the file arguments have patterns to display

1. https://pubs.opengroup.org/onlinepubs/009696799/utilities/what.html

```
%perl what /bin/gprof /bin/bash /bin/echo
/bin/gprof:
         Copyright (c) 1983 Regents of the University of California.
/bin/bash:
        Bash version 5.1.4(1) release GNU
/bin/echo:
```

### DIFF
--- a/bin/what
+++ b/bin/what
@@ -13,33 +13,22 @@ License: perl
 
 use strict;
 
-sub printWhat($$$);
+use Getopt::Std qw(getopts);
 
-my $stop = 0;
-if (($#ARGV >= 0) && ($ARGV[0] eq '-s'))
-{
-    $stop = 1;
-    shift;
-}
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: invalid options die; no longer supports stdin

PR #458, opened 2024-02-13
diff: reduce code noise

mknos:
* Mapping loop is more direct to write with map()
* Check for error instead of success in output_diff()
* Tested this with -e, -f and -u flags on some small files


briandfoy (approved):

briandfoy:
changes: reduce internal clutter

PR #459, opened 2024-02-13
ed: simplify edEdit()

mknos:
* edEdit() handles both "r Filename" and "w Filename"; "w" replaces entire line buffer
* $chars is the number of bytes read into the buffer from the file
* Use $chars directly as the byte count instead of copying the value from $tmp_chars (reduce confusion)
* Variable tmp_lines2 also gets removed
* Return 0 for failure as advertised at top of edEdit()
* Change remaining regex to eq in command interpreter for consistency

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -262,9 +262,9 @@ while (1) {
             &edPrint;
         } elsif ($command eq 'P') {
             edPrompt();
-        } elsif ($command =~ /^q$/) {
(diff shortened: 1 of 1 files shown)


briandfoy (approved):
Good changes.

briandfoy:
changes: improvements to `r filename` and `w filename`

PR #461, opened 2024-02-14
env: print usage for bad options

mknos:
* If I typed the wrong option to env, e.g. -x, print the usage string to show the correct options (BSD version does this)
* pod: clarify that the -u option can be used more than once
* pod: words found by spell-checker


briandfoy (approved):

briandfoy:
changes: print usage for bad options

PR #462, opened 2024-02-14
colrm: accidental $_

mknos:
* In 2-argument mode, the default variable $_ is not used in the loop
* Adding \n back to line is reversing the chomp at start of loop
* I checked that $_ did not have a value here
* test: "perl colrm 2 2 < file.txt" removes the 2nd character on each line of file.txt


briandfoy (approved):

briandfoy:
changes: some internal improvements

PR #463, opened 2024-02-15
od: support multiple input files

mknos:
* Standards document specifies multiple input files should be processed [1]
* Follow BSD od and read stdin only if no arguments are given
* Update usage string
* Options -N and -j affect the input stream overall, not just the 1st file processed
* Uneven input data left over at EOF is combined with input from the next file
* Uneven input data at end of the last file gets its own dump_line() call
* Formatting routines are now written without the need for global $pffmt
* test1: "echo hello |  perl od -c -N 2" --> format 1st two bytes from stdin as characters
* test2: "od -j 1 -x  awk ar" --> dump awk and ar files as hex, skipping 1st byte of input

1. https://pubs.opengroup.org/onlinepubs/009696899/utilities/od.html

### DIFF
--- a/bin/od
+++ b/bin/od
@@ -127,8 +127,7 @@ unless (@ARGV) {
     dump_file(*STDIN);
 }
 dump_line() if (defined $data);
-emit_offset();
-print "\n";
+emit_offset(1);
 exit $rc;
 
 sub limit_reached {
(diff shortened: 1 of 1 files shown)


briandfoy:
`od` has some tests to update to support this change.

mknos:
Sorry about that. I had a look, and the final emit_offset() call was printing a space after the offset, followed by a newline.
Technically the space is not required, only the newline is.

briandfoy:
I see this but I'm too tired this week to understand that change. Give me a couple days to look at it.

briandfoy (changes requested):
I think this is good, but it's a big enough change that I think you should add yourself as an author.

briandfoy (approved):

briandfoy:
changes: support multiple input files

PR #465, opened 2024-02-15
cmp: 2-argument minimum

mknos:
* Usage string states that file1 and file2 are required
* Standards document indicates file2 is not supposed to be inferred as stdin [1]
* Either file1 or file2 can be specified as '-' for comparison against stdin
* test1: error: no args: perl cmp
* test2: error: bad option: perl cmp -x 
* test3: error: file2 omitted: perl cmp test.txt
* test4: error: two stdins: perl cmp - -
* test5: compare stdin with file bc: echo '#!/user/bing' | perl cmp - bc
* test6: skip 1 byte of file1: perl cmp ar awk 1
* test7: skip 1 byte of file1 and file2: perl cmp ar awk 1 1
* test8: error: too many args: perl cmp ar awk 1 1 1

1. https://pubs.opengroup.org/onlinepubs/009604499/utilities/cmp.html

### DIFF
--- a/bin/cmp
+++ b/bin/cmp
@@ -80,19 +80,18 @@ if ($opt{'l'}) {
     $volume = 2;
 }
 $volume = 0 if $opt{'s'};
-usage() unless @ARGV >= 1 and @ARGV <= 3;  # exits;
 
+usage() if (scalar(@ARGV) < 2);
 $file1 = shift;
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: require at least two arguments

PR #466, opened 2024-02-20
ed: edInsert() mostly duplicate code

mknos:
* edInsert() handles Append "a" command and Insert "i" command
* The special case for address 0 can be removed; the code with tmp_lines2 is general enough to handle this, i.e. for address 0, tmp_lines will be added at start of buffer and tmp_lines2 will be empty
* In insert mode the target address is the addressed line; decrement is needed for conversion to correct array index
* test1: "0a", "0i" and "1i" --> add input to start of buffer
* test2: "1a" and "2i" --> add input after buffer line 1, and at buffer line 2 (equivalent)

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -780,37 +780,15 @@ sub edInsert {
         $tmp_chars += length;
     }
 
-    # now that we've got it, figure out what to do with it
+    my $src = $adrs[0];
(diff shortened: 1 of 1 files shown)


briandfoy:
changes: refactor some of the insertion code

PR #467, opened 2024-02-23
tac: strict.pm coverage

mknos:
* The code at beginning of file was not covered by "strict"
* When adding strict at top of file, %opts was not in global scope
* This patch fixes a bug where options such as -s SEP were being ignored


briandfoy (approved):

briandfoy:
changes: fix some odd cases for -s

PR #470, opened 2024-02-26
primes: fail for too many args

mknos:
* This version of primes takes 0, 1 or 2 arguments
* 0 args: read Start from standard input and use default End
* 1 arg: Start is arg1 and End is default
* 2 args: explicit Start and End
* This patch is compatible with NetBSD, which prints usage if >2 args are given

### DIFF
--- a/bin/primes
+++ b/bin/primes
@@ -27,6 +27,10 @@ END {
   $? = 1 if $? == 255;  # from die
 }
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: fail for too many arguments

PR #471, opened 2024-02-26
apply: simplify run_cmd()

mknos:
* It is safe to pass @_ to system() if @_ is empty
* This happens when running apply with the -0 option
* test: "perl apply -0 echo 1 2 3" --> echo runs 3 times with no argument


briandfoy (approved):

briandfoy:
changes: simplify some internals

PR #472, opened 2024-02-26
addbib: -p option handling

mknos:
* If file argument '0' was given to the -p option, it was incorrectly ignored
* extra1: remove warnings.pm as done in other scripts
* extra2: simplify prompt input regex by writing it as a character group; found by perlcritic

### DIFF
--- a/bin/addbib
+++ b/bin/addbib
@@ -13,7 +13,6 @@ License: perl
 
 
 use strict;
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: allow file named 0 for -p

PR #474, opened 2024-02-28
deroff: allow standard input

mknos:
* File argument is documented as optional, but deroff fails when I omit it
* Patch is the result of testing against deroff on OpenBSD
* Convert file processing into a function taking a filehandle param
* Add strict
* Print usage and exit for bad options (previously bad options were treated as a file arg)
* Check return value of close()
* test1: "zcat /usr/share/man/man1/sum.1.gz | perl deroff" --> no args; read stdin
* test2: "perl deroff -" --> error because I have no file called '-' and '-' does not indicate stdin for this program
* test3: "perl deroff deroff deroff" --> 2 file args; program assumes input is in expected format

### DIFF
--- a/bin/deroff
+++ b/bin/deroff
@@ -16,19 +16,49 @@ License: gpl
 # You may use this according to the GNU Public License: see http://www.gnu.org
 # Documentation at bottom.
 
-if ( $ARGV[0] eq '-w' )
-{
-    $Words_Only = 1;
-    shift;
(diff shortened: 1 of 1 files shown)


briandfoy:
I think I understand this patch, but let me check again tomorrow.

briandfoy:
changes: allow reading from standard input

PR #475, opened 2024-02-29
shar: error detected for no files processed

mknos:
* Print usage and exit if no file arguments are given (in docs, at least one file argument is required and standard input is not used)
* Remove variable $dirty for keeping track of whether header was printed; just print it once before loop
* Introduce variable $done to count how many files were processed
* Checking $dirty at end of loop was incorrect; this resulted in printing an error on success and exiting with error code
* test1: "perl shar" --> no args, usage
* test2: "perl shar no-such-file" --> print message "no files were processed"
* test3: "perl shar tar" --> archive one file 'tar' successfully

### DIFF
--- a/bin/shar
+++ b/bin/shar
@@ -31,20 +31,20 @@ sub usage {
 }
 
 getopts('') or usage();
+@ARGV or usage();
 binmode STDOUT;
 
-my $dirty = 0;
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: handle missing or unreadable files

PR #476, opened 2024-03-01
asa: regex too greedy

mknos:
* Problem1: asa would refuse to open a file with a dash
* Problem2: if opening a file fails, asa did not proceed to the next file argument
* This patch brings the code closer to NetBSD version [1]
* test1: "echo hi | perl asa" --> no args; read stdin
* test2: "perl asa real-file" --> one file argument; real-file is not a command option
* test3: "perl asa notfound real-file" --> warn for not opening 'notfound', process 'real-file' then exit with code 1 to indicate error

1.  http://cvsweb.netbsd.org/bsdweb.cgi/src/usr.bin/asa/asa.c?annotate=1.17

### DIFF
--- a/bin/asa
+++ b/bin/asa
@@ -14,24 +14,51 @@ License: perl
 
 use strict;
 
-exit 1 if grep {!-r} @ARGV;		# traditional
+use File::Basename qw(basename);
 
-if (grep /-/, @ARGV) {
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: fix problem reading a file with a dash in name

PR #477, opened 2024-03-01
ping: fatal error for unknown options

mknos:
* Print usage and exit for invalid option -x
* Include -n flag in usage string
* Throw error if extra arguments appear (closer to BSD version)
* test1: "perl ping" --> error; no host
* test2: "perl ping -x" --> error; no -x flag
* test3: "perl ping 127.0.0.1 10" --> success; localhost with timeout of 10 seconds
* test4: "perl ping 8.8.8.8 10 11" --> error; 11 means nothing

### DIFF
--- a/bin/ping
+++ b/bin/ping
@@ -17,12 +17,19 @@ use Getopt::Std;
 use Socket;
 use Net::Ping;
 
-my %opt;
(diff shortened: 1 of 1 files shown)


briandfoy:
changes: reject unknown arguments

PR #478, opened 2024-03-02
uname: reject command arguments

mknos:
* As with bin/arch, uname does not accept any arguments
* For this case the Linux version fails with 'extra operand' error and the OpenBSD version prints usage
* POD after __END__


briandfoy (approved):

briandfoy:
changes: reject any arguments

PR #479, opened 2024-03-02
nl: support one file argument

mknos:
* Standard nl is not required to support multiple input files [1]
* Follow BSD versions of nl and support only one input file, which may be '-' for stdin
* Also checked SunOS manual page, and it is consistent with BSD [2]
* Update usage string and description in POD
* While here, add explicit close() call in do_file() so we can exit with failure code if close() fails
* Delete comment at top of file which duplicates what is written in POD
* Bump version

1. https://pubs.opengroup.org/onlinepubs/009695399/utilities/nl.html
2. https://shrubbery.net/solaris9ab/SUNWaman/hman1/nl.1.html

### DIFF
--- a/bin/nl
+++ b/bin/nl
@@ -2,10 +2,6 @@
 #
 # nl - line numbering filter
 #
-# nl is a clone of the standard 'nl' line numbering utility, in Perl. It reads
-# files sequentially, and writes them to STDOUT, with lines numbered. If file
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: specify exactly one file on the command line

PR #480, opened 2024-03-02
tee: '--' option terminator

mknos:
* The following command works the same on Linux and OpenBSD: "tee -- -mylog"
* The argument list contains one file called "-mylog"; it's not an option as it appears after '--'
* Perl code has a custom option parser so it needs to handle '--' directly


briandfoy (approved):

briandfoy:
changes: end command line options with --

PR #481, opened 2024-03-04
unshar: '--' option terminator

mknos:
* I noticed that shar works with command line "perl shar -- -file1", but unshar doesn't
* There was no unshar command on my Linux or OpenBSD system to test against
* I installed gnu sharutils 4.15.2 and proved that it does support '--'

/usr/bin/unshar -- -file1
fs error 2 (No such file or directory) stat-ing -file1


briandfoy (approved):

briandfoy:
changes: terminate arguments with --

PR #482, opened 2024-03-05
split: raise error for -b 0

mknos:
* "perl split -b 0 file" is invalid but did not raise an error due to truth check on $opt{b}
* -b NUMBER has no default value, but it must be >0
* Add explicit checks for options -b, -l and -p which cannot be used together
* test1: "perl split -b 0 file" --> now fails
* test2: "perl split -b 1 file" --> one byte per file
* test3: "perl split -b 5 file" --> final file with <5 bytes (remainder)
* test4: "perl split -l 2 file" --> two lines per file
* test5: "perl split longfile" --> default -l mode: 1000 lines per file

### DIFF
--- a/bin/split
+++ b/bin/split
@@ -97,6 +97,16 @@ exit 1;
 #### Main program starts here. ####
 
 getopts('b:l:p:', \my %opt) or clue;
+my $nbytes;
+if (defined $opt{'b'}) {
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: disallow a split size of 0

PR #483, opened 2024-03-05
col: buffering code broken

mknos:
* I observed strange output from col when the input file was large
* Finally I discovered the issue is related to code that runs when the buffer fills up
* The input loop adds lines to a buffer which are printed in a subsequent loop; however, the buffer code triggers early printing within the input loop
* I could reproduce error earlier by setting a smaller buffer size with -l:  "perl col -l 5 ching"
* The pod manual says -l buffers at least n lines of input
* I propose ignoring -l and buffering the whole file, this at least doesn't result in a runtime error
* Correct buffer management code could be investigated later

### DIFF
--- a/bin/col
+++ b/bin/col
@@ -121,31 +121,6 @@ while (<>) {
 
         # start row if necessary
         $buf[$row] ||= [];
-
-        # check buffer size
-        if (@buf > $opt_l) {
-            print_line(0);
-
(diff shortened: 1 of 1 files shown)


briandfoy:
Let me think about this one. I'm hesititating because I wouldn't want to buffer an entire file if it's large (hundreds of Mb).  I see that the current state is broken, though.

PR #485, opened 2024-03-06
cut: -f not allowed with -b

mknos:
* Invalid usage resulted in undocumented behaviour; "perl cut -b 1 -f 1" would run handle_b(), exit and ignore "-f 1"
* GNU and OpenBSD versions fail in this case because only one of -b and -f can be specified
* For Perl version, option -c is an alias for -b, so add validation check after equivalence code for -c


briandfoy (approved):

briandfoy:
changes: don't allow -b and -f at the same time

PR #486, opened 2024-03-07
which: 3 distinct exit codes

mknos:
* Previously, bin/which would use exit code 0 (success) even if it failed to locate a command, making it more difficult to use in shell scripts
* Follow GNU version and treat exit code 2 as internal error, 1 as partial-or-no-match, 0 as all-commands-matched
* Follow OpenBSD version and print error to stderr for each command that does not match
* Bump version
* test1: "perl which" --> invalid usage due to no argument, exit(2)
* test2: "perl which sh ls" --> sh and ls are both found on my system, exit(0)
* test3: "perl which sh brash" --> sh is found and brash is not, exit(1)
* test4: "perl which -a awk" --> 3 awks are found on my system, exit(0)

### DIFF
--- a/bin/which
+++ b/bin/which
@@ -17,10 +17,11 @@ use File::Basename qw(basename);
 use Getopt::Std qw(getopts);
 
 use constant EX_SUCCESS => 0;
-use constant EX_FAILURE => 1;
+use constant EX_PARTIAL => 1;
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: use the same exit codes as the GNU version

PR #487, opened 2024-03-08
fmt: indicate open() failure

mknos:
* Explicitly open() input files so we can exit with a failure code if open() fails
* Input loop is converted into a function; line buffer remains global
* Ignore directories as done by GNU fmt
* Read stdin if no file arguments are given
* Tested patch against previous version and I get the same output

### DIFF
--- a/bin/fmt
+++ b/bin/fmt
@@ -42,33 +42,54 @@ format =
 .
 FORMAT
 
-while (<>) {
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: better reporting when the program cannot open a file

PR #488, opened 2024-03-08
sum: ambiguous options

mknos:
* This version of sum attempts to be compatible with NetBSD by offering -o 1 and -o 2 for the historic BSD and SYSV sum algorithms respectively
* When -o is used for selecting an algorithm, forbid -a option which is used for selecting newer algorithms such as md5


briandfoy (approved):

briandfoy:
changes: don't allow -a and -o at the same time

PR #489, opened 2024-03-08
ed: simplify edSetCurrentLine()

mknos:
* Code can be reduced by checking for error instead of checking for success
* Style: consistently return 1 or 0 (return value is not currently checked)
* test1: open file with "perl ed awk", then hit return/enter to proceed to maxline+1 (error for increment case)
* test2: open file with "perl ed awk", then type "1" (success for jump case)
* test3: open file with "perl ed awk", then type "10" then hit extra return (success for increment case after jump to 10)
* test4: open file with "perl ed awk", then type "1000" (error for jump case; line number out of range)

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -860,7 +860,7 @@ sub edQuit {
 sub edSetCurrentLine {
     if (defined($args[0])) {
         edWarn(E_ARGEXT);
-        return;
+        return 0;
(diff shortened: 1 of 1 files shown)


briandfoy:
changes: internal improvements to reduce complexity

PR #490, opened 2024-03-08
glob: handle bare tilde

mknos:
* A tilde by itself signifies $HOME if it is not followed by a username
* I regularly type "cd ~/somedir" but an error was being raised by this version of glob
* Splitting the code into separate regexps makes this a bit clearer
* With this patch the output matches the following globs from bash
* test1: "echo ~" --> /home/pi
* test2: "echo ~/" --> /home/pi/
* test3: "echo ~pi" --> /home/pi
* test4: "echo ~pi/" --> /home/pi/
* This also matches output from perl's glob() on my system


briandfoy (approved):

briandfoy:
changes: handle a bare tilde at the start of the path

PR #491, opened 2024-03-09
expand: single tabstop correction

mknos:
* The default number of spaces for a tab is 8
* The number of spaces is changed by running something like "perl expand -12 file", but when doing this I get one space not 12
* line_desc is only set if multiple tabstops were specified (not relevant to this usage)
* When one tabstop is given, e.g. "expand -12 file", $tabstop has been updated and we should use its value
* Now -12 argument works the same as OpenBSD and GNU version


briandfoy (approved):

briandfoy:
changes: fix the tabstop argument to get the correct number of spaces

PR #492, opened 2024-03-11
expand: '--' option terminator

mknos:
* Accept usage that GNU expand accepts: "perl expand -4 -- -a" for a file called "-a"
* Remove stray quote from error string


briandfoy:
changes: allow for the -- option terminator

PR #493, opened 2024-03-11
tar: file extraction fix

mknos:
* I want to extract a tar
* It's not possible to create a tar with this program because -c option is not implemented
* I create a tar with GNU tar (v1.34), but extraction didn't work
* Problem1: stdin was being read because file descriptor wasn't redirected properly; instead just declare $fh as input file handle
* Problem2: in extract_entry() file mode wasn't being handled properly; shifting bits wasn't correct because mode was already 0755 in the test tar
* Removing the early return allows the tar to be extracted
* test: copy new.tar to a new directory, change directory into it and run "perl $PPT_DIR/tar -x -f new.tar"

### DIFF
--- a/bin/tar
+++ b/bin/tar
@@ -137,23 +137,12 @@ sub extract_entry
    $name =~ s/([A-Z])/_\l$1/g;
   }
  make_dir($1) if ($name =~ m#^(.*)/[^/]+#);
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: fix file extraction

PR #494, opened 2024-03-11
id: don't ignore extra arguments

mknos:
* This version of id follows BSD and only supports one user argument
* Instead of ignoring extra arguments, print usage and exit as done by OpenBSD version
* test1: valid usage: "perl id 0" --> display root user on my Linux system
* test2: invalid usage: "perl id 0 0 0 0 0" --> previously did not raise error
* test3: valid usage: "perl id" --> user taken as "current user"


briandfoy:
changes: allow only 0 or 1 arguments

PR #495, opened 2024-03-11
ar: file header not printed for -q

mknos:
* In quick-append mode (-q), ar creates a file if it doesn't exist and appends to a file if it exists
* I discovered that if a file is created in this way, the valid header record is not printed, making the archive invalid
* GNU ar is smart enough to print file header in this case
* step1: quickly create archive with one file: perl ar -q new.ar addbib 
* step2: append another file to archive: perl ar -q new.ar asa
* step3: list files in archive: perl ar -t new.ar

### DIFF
--- a/bin/ar
+++ b/bin/ar
@@ -16,6 +16,8 @@ use File::Basename;
 use FileHandle;
 use Getopt::Std qw(getopts);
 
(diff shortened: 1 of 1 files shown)


briandfoy:
changes: ensure file  has header with -q, which should have happened before but wasn't

PR #496, opened 2024-03-12
nl: numbering type validation

mknos:
* As outlined in usage string, the valid line numbering types for options -b/-f/-h are a/t/n/pexpr/eexpr
* 1st character of option-value gets split into "type" list; validate this list at start of program
* Invalid usage example: perl nl -b a -f n -h o


briandfoy (approved):

briandfoy:
changes: validate arguments to -b, -f, and -h

PR #500, opened 2024-03-13
ar: warn if archive member isn't found

mknos:
* ar -p mode with no arguments prints all members of the archive to standard output
* ar -p can be given explicit member names to print
* If I type the wrong member name GNU ar gives me a warning and proceeds to print the next member argument
* Example: "perl ar -p new.ar notfound asa" --> print warning on stderr for notfound not being found, then print content of file asa on stdout

### DIFF
--- a/bin/ar
+++ b/bin/ar
@@ -179,10 +179,13 @@ exit 0;
 sub printMember {
     my ($name, $pAr, $verbose) = @_;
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: warn if archive member isn't found

PR #501, opened 2024-03-14
arithmetic: -r validation

mknos:
* Validation code for number range option -r was looking at the wrong variable name
* The intent of the code is for only validated input to be assigned to $range


briandfoy (approved):
Nice catch

briandfoy:
changes: validate the -r argument using the right variable

PR #502, opened 2024-03-14
hangman: more validation

mknos:
* An empty wordlist.txt could be provided, in which case hangman should fail because no lines can be read
* Also, there was no code after the read-words loop to check if $word was given a value
* Rewrite read-words loop without the need for $line_num variable
* test1: create empty wordlist with "touch wordlist.txt"
* test2: create a list with 1 word: "echo hey > wordlist.txt"


briandfoy (approved):

briandfoy:
changes: ensure that we can get at least one word from the word list

PR #503, opened 2024-03-14
printenv: retire -? flag

mknos:
* Remove -? which is not present in GNU or BSD version of printenv
* The BSD version does not support options starting with a dash
* While here, restructure code to avoid multiple exits; exit code 1 is designated for the (optional) argument not being found in environment

### DIFF
--- a/bin/printenv
+++ b/bin/printenv
@@ -14,29 +14,22 @@ License: perl
 
 use strict;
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: remove -? option to match BSD

PR #504, opened 2024-03-14
join: '--' option terminator

mknos:
* This was found when testing against the GNU version of join
* Add handling to custom option parser get_options()
* test: "cp awk ./-a && perl join -- -a -a | md5sum" --> join file "-a" with itself


briandfoy (approved):

briandfoy:
changes: allow -- to terminate options

Issue #505, opened 2024-03-15
Unquoted $^X in date.t


kal247:
I believe it should be `"$^X" bin/date -u`, to avoid CPAN installation failure when $`X contains spaces.

I wish I could make a PR, but I'm completely lost in Git at the moment, sorry.

briandfoy:
It looks like you succeeded in the PR, so I think this issue is complete. If it's not, we can re-open it. Thanks,

PR #506, opened 2024-03-15
fix fox  Unquoted $^X in date.t #505

kal247:
https://github.com/briandfoy/PerlPowerTools/issues/505


briandfoy (approved):
Good catch. Thanks,

briandfoy:
changes: protect path to current perl in backticks (from kal24)

PR #509, opened 2024-03-19
fix caller() for PAR::Packer in modulinos

kal247:


briandfoy (changes requested):
The check for 'main' is a bit weird here and we can't disallow that. That would be surprising to people who write a simple program and want to load one of the modulinos.

Is this something that PAR does that `caller(1)` won't catch?

briandfoy:
Note that this change makes t/factor/factor.t hang.

kal247:
Long story short : 
While I was preparing this PR yesterday, Murphy's law hit me.
Calling the modulinos indirectly, like `bin/perlpowertools echo` (for example) didn't work (anymore?).
So I added this 'main' test to fix the issue, then noticed that the tests for 'factor' and 'units' hang.
At that point I was very confused (doubting of my code, my memory and my two laptops) and decided to start the PR...

kal247:
I'll raise another PR with all the code, so you can have the full picture.

kal247:
Somehow a commit "packer v1.32" has been added to this PR.
The way of Git is really mysterious to me.

kal247:
> The check for 'main' is a bit weird here and we can't disallow that. That would be surprising to people who write a simple program and want to load one of the modulinos.
> 
> Is this something that PAR does that `caller(1)` won't catch?

If `bin/perlpowertools` declared (or reused) a package, would it show as caller(0)? Maybe could we even test `caller(0) eq 'PerlPowerTools'`?

briandfoy:
I don't know anything how packer works, so I don't have any advice to give there. Maybe people on /r/perl or Stackoverflow would know.

kal247:
To fix the issue with 'main' in modulinos : I've set `package PerlPowerTools` before calling `do` in helper script, then changed the test in modulinos.
For 'factor', I had to add `1;` to avoid Failed test 'require 'bin/factor'.
'factor' and 'units' don't hang anymore and `make test` runs ok.

bin/date
```diff
@@ -22,7 +22,8 @@ my $VERSION = '1.0.4';
 # prefer this if it is defined.
 my $TZ;
 
-run(\@ARGV) unless caller();
+# run if called directly, indirectly, directly par-packed, undirectly par-packed
+run(\@ARGV) if !caller() || caller(0) =~ /PerlPowerTools/ || caller(0) eq 'PAR' || caller(1) eq 'PAR';
```
mknos, line 26: I don't understand why this is needed, but caller(0) is being evaluated twice, 1st by regex then by "eq". Would it be better to cover both PAR and PerlPowerTools in the regex?

briandfoy: I like this plan (with some fixes), but let's choose a namespace that only the packer stuff will use just to avoid possible conflicts. Something like PerlPowerTools::Packed that's not already in use.

kal247: Good ideas

kal247: > I don't understand why this is needed, but caller(0) is being evaluated twice, 1st by regex then by "eq". Would it be better to cover both PAR and PerlPowerTools in the regex?

There are some layers and iterations of development on this line of code...

briandfoy:
The latest changes are good. It looks like there's an unrelated files outside of bin/ that snuck in. If we can fix that up I think this request is good to go.

kal247:
Sorry, but where are the unrelated files? I don't find them.

Another question : I've not yet committed the latest 'perlpowertools.exe' file itself. Should I do it, or will it be rebuilt automatically anyway?

Thanks

briandfoy:
When I look at this PR in GitHub, there are 12 files changed. I figure that you merge something and some other files snuck in.

As for rebuilding, I haven't set up anything to do anything with packer. If something needs to be rebuilt we'll have to figure that out. It might happen as part of the release GitHub action.

minitsugu:
The 12 files are changed as intended : 
- 2 changes to move and rename packer.pl -> packer
- 2 new tests
- 1 helper script
- 5 tools fixes for modulinos
- new Readme file
- new binary

kal247:
The 12 files are changed as intended :

    2 changes to move and rename packer.pl -> packer
    2 new tests
    1 helper script
    5 tools fixes for modulinos
    new Readme file
    new binary

briandfoy:
For this PR, you should only be fixing the files under *bin/*. This is the same thing I said before. I'm not going to merge the files outside of *bin/* for this change, so the pull request needs to be fixed.

briandfoy (approved):

kal247:
Thanks!

PR #510, opened 2024-03-20
ar: add strict

mknos:
* ar was mostly compliant with strict except for option variables
* Instead of summing $opt_d, $opt_m etc (which might be undefined), count options using a list
* Don't access $ARGV[0] if it doesn't exist
* test1: "perl ar" --> invalid: need a major option to determine mode
* test2: "perl ar x notfound" --> valid: equivalent to ar -x notfound
* test3: "perl ar -x -m" --> invalid: -m and -x are not compatible

### DIFF
--- a/bin/ar
+++ b/bin/ar
@@ -11,6 +11,8 @@ License: perl
 
 =cut
 
+use strict;
+
 use POSIX qw(strftime);
 use File::Basename;
 use FileHandle;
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: improve internals with strict; better option validation

PR #511, opened 2024-03-20
bc: regression on old perl

mknos:
* The default mode of this bc is not bc at all because it is not arbitrary precision; numbers are just perl floating point values
* It's still useful enough as a basic calculator, but not as a bc-compatible one
* For fun I booted DamnSmallLinux live-cd v4.4.10
* The version of perl on this old linux doesn't include Math::BigFloat
* bc would not load, even in default mode because of a "use" statement; lazy-load with "require" avoids the error
* test1: "perl bc" --> load in not-arbitrary-precision mode and 2**128/33333 is a perl floating point number
* test2: "perl bc -b" --> fail to load Math::BigFloat on old perl

### DIFF
--- a/bin/bc
+++ b/bin/bc
@@ -2034,10 +2034,10 @@ sub command_line()
   while (@ARGV) {
     my $f = shift @ARGV;
     if ($f eq '-b') {
-      require Math::BigFloat;
+      eval { require Math::BigFloat } or die "This program requires the Math::BigFloat module\n";
       $bignum = 1;
     } elsif ($f eq '-d') {
-      require Data::Dumper;
+      eval { require Data::Dumper } or die "This program requires the Data::Dumper module\n";
       $debug = 1;
     } elsif ($f eq '-y') {
       $yydebug = 1;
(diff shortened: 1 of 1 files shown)


briandfoy:
Huh, that's strange since Math::BigInt is part of the core library since Perl 5.

briandfoy (changes requested):
Let's catch failures in the require too.

mknos:
> Huh, that's strange since Math::BigFloat is part of the core library since Perl 5.

I had a feeling stuff had been forcibly deleted from the perl distro on damnsmalllinux 4.4.10 to make the live-cd ISO 50 MB. I just downloaded perl-5.8.0 sources and confirmed that it contains Math::BigFloat version 1.35. The lazy-load patch is still ok but I can add a commit to print a nicer error if Math::BigFloat isn't there.

briandfoy:
That sounds like a good plan. Do that.

briandfoy (approved):

briandfoy:
changes: handle cases where Math::Big* is missing (like minimal distros of perl)

PR #512, opened 2024-03-21
touch: remove -f from usage()

mknos:
* Standards document does not specify behaviour of -f flag for touch, so it's not a bug to leave it unimplemented [1]
* NetBSD and OpenBSD currently treat -f as no-op, but previously OpenBSD implemented it
* No need to document differences from OpenBSD which no longer exist
* Don't suggest -f in usage string because it's silly
* Fix typo in pod

1. https://pubs.opengroup.org/onlinepubs/9699919799/utilities/touch.html

### DIFF
--- a/bin/touch
+++ b/bin/touch
@@ -90,7 +90,7 @@ exit $rc;
 
 sub usage {
     warn "$Program (Perl bin utils) $VERSION\n";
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: de-emphasize -f, which is a no-op anyway

PR #513, opened 2024-03-21
kill: list signals for bad sig name

mknos:
* version1: bash: "builtin kill -h" --> "invalid signal specification" error
* version2: GNU: "/usr/bin/kill -h" --> prints usage string
* version3: pdksh: "builtin kill -h" --> "unknown option" error
* version4: OpenBSD: "/bin/kill -h" --> "unknown signal" error, then print signal list
* To me, version4 is the most helpful because it saves the user having to type "kill -l" to see the list
* test1: "perl kill" --> no pid given; print usage
* test2: "perl kill -h" --> print error, show signal list, exit(1)
* test3: "perl kill -l" --> show signal list, exit(0)
* test4: "perl kill 1" --> fails as expected because my user doesn't have permission to kill init process

### DIFF
--- a/bin/kill
+++ b/bin/kill
@@ -21,27 +21,14 @@ use Config;
 
 # die if no signals or no arguments
 die "No signals defined ?!?" unless defined $Config{"sig_name"};
-die "$0: Too few arguments; try $0 -h\n" unless ( @ARGV );
+usage() unless @ARGV;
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: help shows signal list

PR #515, opened 2024-03-22
du: warn on stat() failure

mknos:
* du can take a number of file arguments
* GNU and OpenBSD versions will print a warning if a file fails to stat, then proceed to the next file
* test: "touch there1 there2 && perl du nothere1 nothere2 there1 there2" --> two warning lines and two du summary lines


briandfoy (approved):

briandfoy:
changes: warn if du cannot stat a file

PR #516, opened 2024-03-23
maze: options before arguments

mknos:
* The usage string is the source of truth; optional maze algorithm flag goes before optional width and height arguments
* Remove the grep that reconstructs ARGV with options removed; this caused the incorrect usage "perl maze 20 30 -sf" to be accepted


bin/maze
```diff
@@ -34,26 +34,20 @@ sub traverse_randomly         { rand(@walk) }   # fiendish mazez (random walks)
 sub traverse_randomly_deep    {-rand(@walk/2) } # longer random walks
 sub traverse_randomly_shallow { rand(@walk/2) } # shorter random walks
 
+my %alg = (
```
briandfoy, line 37: Nice change there. I like it.

briandfoy (approved):

briandfoy:
changes: clarify argument processing

PR #517, opened 2024-03-24
bc: assign '=' should not print

mknos:
* Previously this change was committed for internal =P instruction, which runs when assigning an array value, e.g. "ary[0] = 1"
* Apply the same change to the =V instruction, for regular variable assignments
* Instruction =V happens for shortcut assignment operators too
* This patch makes bc agree with GNU version and Gavin Howard version

x = 1
x += 2
x -= 1
x *= 4
x /= 2
x ^= 2


briandfoy:
changes: assignment to variable does not print, in line with GNU bc

PR #521, opened 2024-03-25
bc: forbid negative array index

mknos:
* Negative array index is being caught when accessing a value (p instruction), but is not caught when assigning a value (P instruction)
* Adapt existing guard code from 'p' instruction
* For consistency with undefined-function error, replace YYERROR call with "last INSTR"
* Only the 2nd-from-top ope_stack element is needed; fetch it with splice instead of pop+push dance for $value

```
%perl bc -d # before patch
ary[-1]
instruction: N, 1
instruction: m_
instruction: p, ary
Non-integer index -1 for array

ary[-1] = 123
instruction: N, 1
instruction: m_
instruction: N, 123
instruction: P, ary
instruction: =P
wat
```

### DIFF
--- a/bin/bc
+++ b/bin/bc
@@ -2459,10 +2459,14 @@ sub exec_stmt
      next INSTR;
 
    } elsif($_ eq 'P') {
-
-	 my $value = pop @ope_stack;
-	 my $index = pop @ope_stack;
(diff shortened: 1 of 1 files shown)


briandfoy:
changes: catch negative index in array assignment

PR #526, opened 2024-04-02
printf: remove eval

mknos:
* Rewrite printf command to not use eval()
* Format string is parsed into a list of literal substrings and format specifiers
* The print loop executes at least once; missing argument are implied as 0 for numbers and '' for strings
* Remove RESTRICTIONS section from pod now that we check input and don't allow arbitrary things into eval()
* Previous version could be exploited like this: perl printf '";;CORE::dump();";dumpassaasdasd'
Aborted

* test1: perl printf 'hey%d' --> implicit argument of zero, "hey0"
* test2: perl printf 'hey%d' 1 --> explicit argument, "hey1"
* test3: perl printf 'hey%d' 1 2 3 --> format string applied 3 times, "hey1hey2hey3"
* test4: perl printf '%20s:%-20s' pri ntf --> length specifier with string
```
                 pri:ntf                 
```
* test5: perl printf --> error, at least the format string is required
* test6: perl printf "\tX\r\n" | xxd --> c-escapes produce hex output: 0958 0d0a
* test7: perl printf '' --> empty format string is not an error; no output

### DIFF
--- a/bin/printf
+++ b/bin/printf
@@ -28,16 +28,66 @@ unless (@ARGV) {
     exit EX_FAILURE;
 }
 
+my @fmt;
 my $format = shift;
-$format =~ s/\\v/\x0b/g; # escape \v not available in printf()
-$format =~ s/\%c/\%\.1s/g; # standard printf: %c == 1st char
-
-my @ints = map { m/\A0x/i ? hex : int } @ARGV;
-eval qq(printf "$format", \@ints) or do {
-    warn "$Program: $@\n";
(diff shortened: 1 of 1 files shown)


briandfoy (approved):
Wow, that eval STRING is so unnecessary. That's a thing we test in Learning Perl. People have this weird idea that the first argument—the format—has to be a literal. Then stuff like this happens.

briandfoy:
changes: remove internal string eval to allow more things in formats and args

PR #527, opened 2024-04-02
spell: compatible -d option

mknos:
* On OpenBSD and NetBSD I run "spell -d PATH" if I want to use a custom dictionary [1] [2]
* FreeBSD and solaris9 do not have the -d option, but define -b as a flag (not option with value) meaning to use the british dictionary not the american one [3] [4]
* Remove -b instead of re-defining its meaning; introduce -d instead
* Allow a space between -d and the file, as supported by existing -s option
* Update pod text and usage string

1. http://man.openbsd.org/spell
2. https://man.netbsd.org/spell.1
3. https://man.freebsd.org/cgi/man.cgi?query=spell&apropos=0&sektion=0&manpath=FreeBSD+14.0-RELEASE+and+Ports&arch=default&format=html
4. https://shrubbery.net/solaris9ab/SUNWaman/hman1/spell.1.html

### DIFF
--- a/bin/spell
+++ b/bin/spell
@@ -36,7 +36,7 @@ my (
 );
 
 sub usage {
-  warn "usage: $Program [-b[dict]] [-c|-x] [-v] [-i] [-s dict] [file ...]\n";
+  warn "usage: $Program [-d dict] [-c|-x] [-v] [-i] [-s dict] [file ...]\n";
   exit EX_FAILURE;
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: remove -b option, use -d instead to use alternate dictionary

PR #529, opened 2024-04-03
tac: validate input for -S option

mknos:
* tac -S option takes a number of bytes to read; this option does not exist in GNU version
* It doesn't make sense for -S value to be negative or zero so enforce a minimum of 1
* Printing usage string seemed slightly helpful, so create a usage() function
* The -S option might provide no benefit to the user and could be removed in future

### DIFF
--- a/bin/tac
+++ b/bin/tac
@@ -28,11 +28,7 @@ my $Program = basename($0);
 my $VERSION = '0.18';
 
 my %opts;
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: validate the value for -S; must be positive

PR #530, opened 2024-04-03
spell: fail for empty dictionary

mknos:
* The dictionary file is read into memory before taking input list of words to check
* Fail early in extreme case where no words were read from dictionary, e.g. touch empty && spell -d empty
* Avoid open() on a directory, e.g. spell -d ..


briandfoy (approved):

briandfoy:
changes: fail if dictionary is empty

PR #531, opened 2024-04-03
whois: fail for bad options & no args

mknos:
* Print usage string for incorrect options; previously invalid option -X was treated as a domain name
* Write "\r\n" to socket, not "\n\r" (hint taken from NetBSD code)
* Don't connect to whois server if no domain arguments were provided; print usage string in this case
* Make the different whois server options mutually exclusive 
* Test: I still get a positive response for "perl whois installgentoo.com"

### DIFF
--- a/bin/whois
+++ b/bin/whois
@@ -20,20 +20,24 @@ License:
 # Added a -6 switch for 6BONE (whois.6bone.net)
 # Added a -g switch for .gov (whois.nic.gov)
 
+use Getopt::Std qw(getopts);
 use IO::Socket;
 
-my $host = "whois.internic.net";
-my $i;
-while($i = shift) {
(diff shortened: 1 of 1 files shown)


bin/whois
```diff
@@ -20,20 +20,24 @@ License:
...
-    else               { unshift(@ARGV, $i);        last; }
-}
+use constant EX_SUCCESS => 0;
+use constant EX_FAILURE => 1;
+
+my (%opt, $host);
+getopts('6adprgh:', \%opt) or usage();
+usage() unless @ARGV;
+
+set_host('whois.arin.net')     if $opt{'a'};
```
briandfoy, line 33: a hash would work well here, but this is fine.

briandfoy (approved):

PR #532, opened 2024-04-04
printf: ignore args case

mknos:
* I discovered incorrect looping behaviour for: perl printf a b c
* In the loop, b and c are never shifted off ARGV list, so it never terminates
* The 1st argument is interpreted as the format string, but it contains no format specifiers
* bash-builtin printf, as well as GNU and OpenBSD versions agree that arguments b and c should not be used
* On OpenBSD there is no pdksh builtin printf to test against
* Text in the standards document is not very helpful: "If the format operand contains no conversion specifications and argument operands are present, the results are unspecified." [1]

1. https://pubs.opengroup.org/onlinepubs/9699919799/utilities/printf.html

### DIFF
--- a/bin/printf
+++ b/bin/printf
@@ -31,7 +31,7 @@ unless (@ARGV) {
 my @fmt;
 my $format = shift;
 exit EX_SUCCESS unless (length $format);
-parse_fmt();
+@ARGV = () if (parse_fmt() == 0);
 do {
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: ignore arguments if the first argument does not have format specifiers

PR #533, opened 2024-04-04
whois: submit new query per domain

mknos:
* NetBSD version of whois takes multiple domain names
* NetBSD version does a new connection for each domain name, so we can follow this [1]
* The query submitted by this version is not correct when multiple names are specified
* With this patch I can successfully query 2 domains in one go

1. http://cvsweb.netbsd.org/bsdweb.cgi/src/usr.bin/whois/whois.c?annotate=1.36.30.1

```
%perl whois installgentoo.com omledom.com   # before patch
No match for "INSTALLGENTOO.COM OMLEDOM.COM".
```

### DIFF
--- a/bin/whois
+++ b/bin/whois
@@ -41,25 +41,28 @@ set_host('whois.internic.net') unless defined $host;
 
 $| = 1;
 
-my $sock = IO::Socket::INET->new(
-    PeerAddr => $host,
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: query multiple domains in one command

PR #534, opened 2024-04-04
kill: validate pid

mknos:
* Avoid passing non-numeric pids to kill()
* Printing "Can't kill a non-numeric process ID" is not helpful because kill can accept many pids and the error doesn't indicate which pid is bad; print a helpful error instead
* This was found when testing against the bash-builtin kill


briandfoy (approved):

briandfoy:
changes: pid must be a positive whole number

PR #535, opened 2024-04-04
comm: don't enforce grouped options

mknos:
* GNU and OpenBSD versions of comm accept this usage: "comm -1 -2 -3 -- -file1 -file2"
* This version only worked if options were grouped as -12 or -123
* Conversion to getopt allows grouped and ungrouped usage, and supports '--' terminator (useful for files starting with dash)
* COL list controls which columns are suppressed, so $COL[1] is false if -1 is given, etc

### DIFF
--- a/bin/comm
+++ b/bin/comm
@@ -22,25 +22,18 @@ License: public domain
 use strict;
 
 use File::Basename qw(basename);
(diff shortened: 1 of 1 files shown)


briandfoy:
changes: allow for the unbundling of options

PR #536, opened 2024-04-05
mail: connect() error tweak

mknos:
* When sending a mail message we need to connect to a relay host
* Append the system error string to the nice error in $emesg, which is already terminated with a newline
* This makes the error take up 2 lines but I think it's still better because we can see more specific error info
```
%RELAYHOST=127.0.0.1 perl mail go@google.com # now shows...
Unable to connect to the specified relay host
IO::Socket::INET: connect: Connection refused
```

### DIFF
--- a/bin/mail
+++ b/bin/mail
@@ -112,9 +112,7 @@ sub new {
 			Type => SOCK_STREAM,
 			Timeout => 15,
 	);
-	if (! $socket) {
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: use error from IO::Socket if there's a problem

PR #538, opened 2024-04-08
whois: forbid query containing only whitespace

mknos:
* A correctly formed whois query will contain non-whitespace characters so catch invalid input before connecting to whois server
* Example bad input: perl whois ' '


briandfoy (approved):

briandfoy:
changes: hostname must have at least one non-whitespace character

PR #540, opened 2024-04-09
bc: unifdef YYDEBUG

mknos:
* Remove #ifdef checks around $yydebug usage
* Toggle parser debugging by running "bc -y"
* Follow the pod text and don't look at environment variable YYDEBUG to determine whether to run in -y mode
* Toggle output state machine debugging by running "bc -d"
* Some statements run under "bc -d" need Dumper() imported into namespace; previous commit broke that
* Simplify variable init in yyparse(); write $yyss[$yyssp] as $yyss[0]

### DIFF
--- a/bin/bc
+++ b/bin/bc
@@ -822,8 +822,6 @@ sub debug(&) {
     if $debug;
 }
 
-;#$yydebug=1;
-
 #line 32 "y.tab.pl"
 $INT=257;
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: internal fixes to make the code simpler

PR #541, opened 2024-04-09
arithmetic: empty answer is not a number

mknos:
* regex m/\D/ didn't match on empty string, so previously if I typed enter instead of a number it was treated as a wrong answer
* It's more correct to treat empty line as bad input and prompt again for a number
* While here, make the input loop more user friendly by trimming leading and trailing spaces before testing for m/\D/

### DIFF
--- a/bin/arithmetic
+++ b/bin/arithmetic
@@ -14,7 +14,6 @@ License: perl
 
 use strict;
 use integer;
-use warnings;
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: treat an empty line as bad input instead of the number undef

PR #542, opened 2024-04-10
ed: input filename validation

mknos:
* An empty filename (ed '') is incorrect usage and should show an error
* GNU ed attempts open() and fails; OpenBSD ed fails with "invalid filename" error and avoids open() (follow OpenBSD)
* In edEdit(), $filename could still be undefined in default case because $RememberedFilename is undefined when starting ed with empty buffer
* The code was a bit redundant because the value of $RememberedFilename was saved into itself again
* Save $RememberedFilename only if filename is taken from $args[0]


briandfoy (approved):

briandfoy:
changes: filename must have at least one character; better support to remember filename

PR #543, opened 2024-04-11
bc: avoid extra copy of filehandle

mknos:
* In next_file(), the old filehandle $input is closed if we're not processing the 1st file
* The filehandle $input is global; open it directly instead of copying into it from "IN"
* IN was not referenced anywhere else in the program


briandfoy (approved):
heh, that's one of those weird coding things where the author was probably thinking of something else and never came back to clean this up. We have all been there.

PR #547, opened 2024-04-11
tail: die for bad options

mknos:
* GetOptions() return value was not checked
* Symbolic exit codes were already declared so use them more consistently -- this usage() takes an exit code parameter


briandfoy:
changes: exit with error if the command line options are wrong

PR #551, opened 2024-04-13
fish: fold case

mknos:
* I am lazy to type shift+a for ace; do what I expect whether I type A or a
* This could've been done in less code, but I wanted a table of valid card names instead of using grep()
* Catch^C^Cgoing fish is so much more fun now


briandfoy (approved):

briandfoy:
changes: card names are now case insensitive

PR #552, opened 2024-04-14
mkdir: make empty mode an error

mknos:
* Follow GNU & OpenBSD mkdir by failing early if the mode string (-m option-argument) was empty
* Previously -m '' was interpreted as -m 0
* Make one error message easier to read


briandfoy (approved):

briandfoy:
changes: die for an empty mode argument

Issue #522, opened 2024-03-25
bc statements have stopped working


briandfoy:
I didn't bisect to find the problem, but this started breaking with #517. I caught up the rest of the tests for that change, but these are breaking for other reasons. Maybe the `print` is broken:

```
$ bc -e 'for ( v=0; v<5; v++) { print v; if (v>2) break }; print "\n"'
0123

$ perl bin/bc
for ( v=0; v<5; v++) { print v; if (v>2) break }; print "\n"
0
```

@mknos can you look at this? I can make these tests TODO, but I'd rather fix them so we can make a release.

mknos:
Thanks for the notification. The commit added for =v instruction caused the loop to terminate because $return=1 breaks out of a group statement. Setting $return=3 as done in FOR-COND fixes this for a basic loop.
However, the example loop with "break" incorrectly printed 01230 with older version of bc. The extra 0 was added by "break". Setting $return=1 as done in RETURN fixes this.
I've added these changes as separate commits in PR523.
It would be good to create some symbolic constants with helpful names for these $return values instead of 0, 1, 2, 3.

briandfoy:
Noting #523 for the cross-reference. (Type # followed by number to refer to another issue in the same repo)

PR #554, opened 2024-04-15
bc: add strict

mknos:
* In yyrule list, strings were not escaped properly
* Use yyclearin() where possible to reset $yychar
* Eliminate #line directives which cause perl debugger to print misleading line numbers
* In grammar rule 84 (expr=IDENT), access $yyvs[$yyvsp] directly instead of accessing a hash inside it which doesn't exist (this is consistent with rules 86,87,etc
* Replace string-eval with block-eval in main()

### DIFF
--- a/bin/bc
+++ b/bin/bc
@@ -801,11 +801,12 @@ repository:
 =cut
 
 #define YYBYACC 1
-#line 49 "bc.y"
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
We're still blocking on #522

PR #556, opened 2024-04-15
patch: refuse to process a directory

mknos:
* When testing against GNU patch, the following usage is invalid: "patch . < some.diff"
* An error is displayed: "File . is not a regular file -- refusing to patch"
* Even with patch -f (force) option the usage is not allowed
* Add extra branch to condition where file-exists check happens to handle this

### DIFF
--- a/bin/patch
+++ b/bin/patch
@@ -382,7 +382,9 @@ sub bless {
     }
 
     # Make sure original file exists.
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: only process regular files, and die otherwise

PR #557, opened 2024-04-15
awk: now with File::Temp

mknos:
* Take a hint from bin/find program, and use File::Temp instead of manually opening TMPIN and TMPOUT
* File::Temp also handles removal of the files, so the END block gets simplified
* Brace filehandle argument in one instance of print()
* "perl awk -f hello.awk" still produced the same output with patch applied

### DIFF
--- a/bin/awk
+++ b/bin/awk
@@ -17,17 +17,15 @@ License: perl
 
 use strict;
 
+use File::Temp qw(tempfile);
+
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: use File::Temp instead of rolling our own

PR #563, opened 2024-04-17
touch: -r argument truth check

mknos:
* Option -r takes a file argument to copy stat() data from
* Fix issue where filename of "0" is ignored because perl treated it as no-r-argument-given


briandfoy (approved):

briandfoy:
changes: -r can now take a filename named "0"

PR #564, opened 2024-04-17
printf: descriptive error message

mknos:
* Follow GNU printf command and attempt to show the bad format specifier which caused an error
* Example: "perl printf '%9 -asd'" now reports "printf: invalid format: '%9'" instead of just "internal error"


briandfoy (approved):

briandfoy:
changes: better error message for bad printf format

PR #565, opened 2024-04-17
shar: handle filename with space

mknos:
* shar from GNU sharutils can work with a filename with spaces, but this version can't
* Address this by quoting the filename passed to mkdir and sed
* It's not harmful to quote files which don't have a space
* While here, print an error if close() failed

```
%perl shar '0 1' > new.shar
%sh new.shar
x - 0 1
sed: can't read 1: No such file or directory
```

### DIFF
--- a/bin/shar
+++ b/bin/shar
@@ -43,7 +43,7 @@ my $done = 0;
 ARGUMENT: for my $f ( @ARGV ) {
     if (-d $f) {
 	print "echo x - $f/\n";
-	print "mkdir -p $f\n";
+	print "mkdir -p '$f'\n";
 	$done++;
 	next ARGUMENT;
     }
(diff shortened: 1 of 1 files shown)


bin/shar
```diff
@@ -43,7 +43,7 @@ my $done = 0;
 ARGUMENT: for my $f ( @ARGV ) {
     if (-d $f) {
 	print "echo x - $f/\n";
-	print "mkdir -p $f\n";
+	print "mkdir -p '$f'\n";
```
briandfoy, line 46: what is the filename has a ' in it though?

briandfoy (changes requested):
Can we fix up the problem of the filename having the quote character as a literal character? That would have already been a problem, but now that we're here maybe we can handle that.

mknos:
Being paranoid here is good. I committed a version that now works with filenames containing a quote.

briandfoy (approved):

PR #567, opened 2024-04-18
cmp: remove -?

mknos:
* Non-standard -? option doesn't exist in BSD versions of cmp either
* For systems with no perldoc command installed we provide a bin/perldoc, so running "perl perldoc cmp" is preferred (i.e. the same way to show the manual for maze and other programs too)


briandfoy (approved):

briandfoy:
changes: remove -? option, although using it give you the help still

PR #568, opened 2024-04-19
rmdir: stat file first

mknos:
* Avoid calling rmdir() on file arguments unless stat (via -d ARG) indicates the file exists and is a directory
* Technically rmdir() would fail anyway, but this seems more defensive
* Symbolic exit codes were already declared here, so use them


briandfoy (approved):

briandfoy:
changes: don't rmdir() if the argument is not a directory

PR #569, opened 2024-04-19
moo: usage() > warn("Usage")

mknos:
* Introduce a regular usage() function
* Size of 0 is still not allowed (the puzzle must have 1 or more numbers)


briandfoy (approved):

briandfoy:
changes: argument must be greater than zero

PR #571, opened 2024-04-22
file: avoid exit(0) for stdin case

mknos:
* The usage "file -" is allowed for GNU and OpenBSD: explicitly read stdin
* The usage "file" with no argument is not valid; stdin is not read implicitly
* This version doesn't support "file -", but the argument loop terminates and results in exit(0)
* Using die for this not-implemented error results in non-zero exit code, which is more compatible with the standard[1]

1. https://pubs.opengroup.org/onlinepubs/007904975/utilities/file.html


briandfoy (changes requested):
Let's exist with something that isn't 255.

briandfoy (approved):
Thanks,

PR #577, opened 2024-04-24
ed: runtime error in edSubstitute()

mknos:
* The program crashes when fed unexpected input after the initial "s" command
* I found this when investigating a capture-variable-outside-of-condition warning from perlcritic
* $char is the 1st character after the "s"; normally it would be "/" as in s/old/new/
* Return early if $args[0] is an empty string
* Adding eval guard around the confusing regex constructed from $char prevents the error
* Possibly the code could be changed later to enforce that $char is '/'
```
perl ed -p 'ed% ' a.c
63
ed% 1
#include <stdio.h>
ed% s*asaaa*
Unknown verb pattern '' in regex; marked by <-- HERE in m/((*) <-- HERE [^"*"]*(*)[^"*"]*(*)?)([imsx]*)/ at ed line 519, <> line 2.
```

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -506,7 +506,7 @@ sub edSubstitute {
         edWarn(E_ADDRBAD);
         return;
     }
-    unless (defined($args[0])) {
+    if (!defined($args[0]) || length($args[0]) == 0) {
         edWarn(E_PATTERN);
         return;
     }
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: handle argument to -s being zero length

PR #578, opened 2024-04-25
diff: empty files are the same

mknos:
* Test case: "touch empty && perl diff empty empty"
* Program seems to hang; ctrl-c in perl debugger hints that we are stuck in Algorithm::Diff::traverse_sequences()
* Program grew to use >2GB of memory before I killed it
* traverse_sequences() doesn't seem to handle when both lists of input lines are empty (I didn't bother investigating why)
* Put a sanity check before diff()+traverse_sequences() are called; input files are read in total before diff() so the size of line list is know
* I also tested a non-empty file against an empty one

### DIFF
--- a/bin/diff
+++ b/bin/diff
@@ -145,6 +145,7 @@ chomp(@f1 = <$fh1>);
 close $fh1;
 chomp(@f2 = <$fh2>);
 close $fh2;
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: empty files are the same

PR #579, opened 2024-04-26
ed: no-match error for s///

mknos:
* Difference observed when testing against GNU ed and OpenBSD ed
* If current line doesn't contain "A", the command s/A/B is expected to generate the error "no match"
* This error happens for searches (? and /), but not for s///
* The logic in edSubstitute() was wrong because LastMatch assumed the current line was a positive match before the loop starts
* The corrected version sets LastMatch only if a match occurred

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -530,9 +530,6 @@ sub edSubstitute {
     }
 
     # do the search and substitution
-
-    $LastMatch = $CurrentLineNum;
-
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: s/// correctly reports "no match" for no matches.

PR #580, opened 2024-04-26
bc: divide by 0

mknos:
* Catch illegal division before it happens
* In bignum mode an error would not be thrown previously for this case; testing with is_zero() resolves this
* I tested this in interactive mode for -b and not-b; the die() is caught so bc does not terminate---this is consistent with GNU bc


briandfoy (approved):

briandfoy:
changes: catch divide by zero before we try the division

PR #582, opened 2024-04-28
bc: symbol table correction

mknos:
* Entries in sym_table hash need to have type & value
* When investigating a warning for "eq on undefined value" on L2400 I noticed that the newly created sym_table entry needs a type of "var" (this code handles auto-assignment of zero to variables)
* The hint came from further down in the code, where array values are initialised to zero (virtual 'p' instruction)
* Also remove the C preprocessor declaration of YYBYACC which was not referenced anywhere

### DIFF
--- a/bin/bc
+++ b/bin/bc
@@ -800,8 +800,6 @@ repository:
 
 =cut
 
-#define YYBYACC 1
-
 use strict;
 
 # The symbol table : the keys are the identifiers, the value is in the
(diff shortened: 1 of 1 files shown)


briandfoy:
This seems to fix #522

briandfoy:
changes: fix problem where vars did not have a type in the internal symbol table (fixes #522)

PR #584, opened 2024-04-30
bc: standard array index

mknos:
* Variables in bc are can have a fractional part
* A variable can be used as an array index
* Discard the fractional part of index, so 3.3 is equivalent to 3
* Negative array index is still not allowed
```
%perl bc 
>>> ary[3.3] = 2
>>> ary[3.3]
2
>>> ary[3]
2
>>> a = 3.4 
>>> ary[a]
2
quit
```


briandfoy (approved):

briandfoy:
changes: allow array indices to have fractional parts (still only use the floor)

PR #587, opened 2024-04-30
pig: support --version

mknos:
* -v flag was mentioned in usage string but not in SYNOPSIS
* Follow what was done in bin/rev: remove -v and support --version by declaring VERSION_MESSAGE
* Now usage string and SYNOPSIS agree (--version is not explicitly mentioned)


briandfoy:
changes: remove -v but support --version

PR #589, opened 2024-04-30
units: remove nested GPL document

mknos:
* --warranty and --copying are no longer options; if this was a good idea it would be included in every single perl script and not just this one
* Replacing the statement before END marker prevents the warning "Useless use of a constant ("PerlPowerTools::units") in void context"
* Fix a typo in pod manual


briandfoy (approved):
We already distribute this license too

briandfoy:
changes: remove --copying and --warranty options (license is in this distro)

PR #593, opened 2024-05-01
nl: retire -V

mknos:
* Linux and OpenBSD don't support nl -V for displaying version
* Standards document doesn't specify a version flag[1]
* GNU nl has --version, which can be supported here as done for bin/rev

1. https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/utilities/nl.html


briandfoy (approved):

briandfoy:
changes: remove -V in favor of --version

PR #594, opened 2024-05-01
xargs: eof ignored

mknos:
* Test cases: "perl xargs" and "perl xargs ls"
* Text entered manually:
1
2
3
^D
* xargs correctly runs a program (echo or ls) on 1, 2 and 3, then incorrectly waits to read more data from stdin
* Checking eof() after each read indicates whether xargs should terminate (no more args will follow so the program should not loop)
* Tweak usage string to clarify that "prog" is optional (echo is the default)


briandfoy:
changes: check eof() to know when to stop

PR #595, opened 2024-05-03
cut: fix substring for -b flag

mknos:
* The substring length calculation in handle_b() wasn't correct in all cases
* If start==end, then length==1
* Pass parameters to handle_b() and handle_f()
* Use defined() to test for presence of command options; 0 should not be ignored
* test1: "perl cut -b 2-3 a.c" -> chars 2-3
* test2: "perl cut -b -3 a.c" -> chars 1-3
* test3: "perl cut -b 10 a.c" -> char 10
* test4: "perl cut -f 1,2 -s  -d ' ' a.c" -> fields 1&2, delimited by space, ignoring lines without a space


briandfoy:
changes: ensure -b returns the right range of characters

PR #597, opened 2024-05-10
pom: avoid using bad environment values

mknos:
* Deliberate crash1: COLS=-1 perl pom -e Out of memory!
* Deliberate crash2: LINES=-1 perl pom -e Can't take sqrt of -0.5625 at pom line 298.
* For crash1 I noted the value of $terminal_width was 4294967292, i.e. sign-extension of -1
* Avoid these errors by displaying a "bad terminal width" or "bad terminal height" error instead
* Make use of checknum() which is already able to catch negative numbers, and other bad input
* Extra change: convert aspect ratio to a constant

### DIFF
--- a/bin/pom
+++ b/bin/pom
@@ -247,9 +247,15 @@ if ($debugging) {
 ## Draw the moon.
 
 my $terminal_width = ($ENV{COLS} || $ENV{COLUMNS} || COLS_DEFAULT) - 2;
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: ensure values for columns, rows, and terminal width make sense

PR #598, opened 2024-05-10
ed: refactor command tree

mknos:
* Replace large if-else block with a more declarative lookup table
* Some helper functions were needed for functions which implemented multiple commands (e.g. edHelp() implements 'h' and 'H')
* Add missing default "else", i.e. print a warning if edParse() was successful but the handler didn't exist in cmdtab hash
* No intended functional change, aside from the extra bad-command warning

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -133,6 +133,33 @@ my %WANT_FILE = (
     'W' => 1,
 );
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):
This is a very pleasing change. Thanks,

PR #599, opened 2024-05-11
factor: pod update

mknos:
* Numbers with leading '-' are not accepted by this version
* I tested GNU factor and it didn't accept negative numbers either
* Text about BSD manual being incorrect might have been relevant 20 years ago, but current OpenBSD manual states that numbers may only have a leading '+' [1]

1. http://man.openbsd.org/factor


briandfoy (approved):

briandfoy:
changes: docs note that this version fixes a bug in the BSD version

PR #600, opened 2024-05-11
bc: remove alias sub

mknos:
* YYERROR() was an alias for yy_err_recover()
* YYERROR() was only called once, so call yy_err_recover() directly
* When testing, I temporarily forced $return=2 instead of working out how to trigger it


briandfoy (approved):
This is a good change and I hope we get more of these sorts of trimmings in bc

PR #601, opened 2024-05-13
wc: indicate error in exitcode

mknos:
* Follow GNU and OpenBSD versions by doing non-zero exit() if one or more file arguments cannot be processed
* Add return check for close() as done in other scripts
```
%echo $BASH_VERSION
5.1.4(1)-release
%set -e
%perl wc -l not-exist
```

### DIFF
--- a/bin/wc
+++ b/bin/wc
@@ -198,19 +198,25 @@ sub wc_fh {
     print "$out";
 }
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):
Nice.

briandfoy:
changes: failures exit with non-zero now

PR #607, opened 2024-05-17
grep: pattern of 0

mknos:
* The following correct usage was not accepted: "perl grep 0 file"
* Within parse_args(), $opt{e} was initialised to 0 but that is confusing because 0 could be a pattern; instead set it to empty string
* First check if $opt{e} and use it if not empty
* Expect pattern to be the next ARGV element
* Raise error via usage() only if there were no remaining ARGV elements (previously usage() was called if it was 0, i.e. false)

### DIFF
--- a/bin/grep
+++ b/bin/grep
@@ -136,8 +136,8 @@ sub parse_args {
 
 	$optstring = "incCwsxvhe:f:l1HurtpP:aqTF";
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: the default for -e is now the empty string, and the value 0 is taken to be a valid pattern

PR #608, opened 2024-05-18
patch: single definition of end()

mknos:
* Make the regular Patch::end() aware that print_tail() should not be called for ed diffs
* Now the subclass definition of Patch::Ed::end() can go away

### DIFF
--- a/bin/patch
+++ b/bin/patch
@@ -504,12 +504,10 @@ sub error {
 # End of patch clean up.
 sub end {
     my $self = shift;
-
-    return if ref $self eq 'Patch';
-
+    my $class = ref $self;
+    return if $class eq 'Patch';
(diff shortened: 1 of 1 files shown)


bin/patch
```diff
@@ -504,12 +504,10 @@ sub error {
 # End of patch clean up.
 sub end {
     my $self = shift;
-
-    return if ref $self eq 'Patch';
-
+    my $class = ref $self;
+    return if $class eq 'Patch';
```
briandfoy, line 508: Should this be an isa check or is that too pedantic?

briandfoy (changes requested):
Just a question. The answer might be that this change is fine.

briandfoy (approved):

PR #609, opened 2024-05-20
cp: print usage for bad options

mknos:
* Preserve control structure of returning from run() by passing NOEXIT to pod2usage()
* Remove empty line in SYNOPSIS and Bquote which didn't look right in the pod2usage() output
* Tested for "perl cp -x" and "perl cp src"


briandfoy (approved):

briandfoy:
changes: print usage and exit for bad options

PR #610, opened 2024-05-20
lock: usage() exits

mknos:
* Reduce code by making usage() not return; this convention is followed by other scripts
* While here, correct spelling errors in pod


briandfoy (approved):

briandfoy:
changes: fix some doc typos

PR #611, opened 2024-05-20
ar: sanity check archive argument

mknos:
* Don't attempt to open an archive file if no archive was specified (all valid usages require an archive to operate on)
* exit(1) in usage() to indicate that something went wrong (matches GNU version)


briandfoy (approved):

briandfoy:
changes: check that there's a defined archive argument before proceeding

PR #612, opened 2024-05-20
rm: print usage for bad options

mknos:
* Don't ignore bad options
* e.g. for usage "perl rm -x f1", file f1 should not be removed
* Reduce verbose level of pod2usage() so only SYNOPSIS and Options are displayed
* I found this when testing against GNU rm


briandfoy (approved):

briandfoy:
changes: fail and print usage message for unsupported options

PR #614, opened 2024-05-21
units: correct usage string

mknos:
* Usage in SYNOPSIS was correct
* usage1: no command args; all input is entered after prompts
* usage2: two unit names are given as arguments, e.g. "perl units cm inch"


briandfoy (approved):

briandfoy:
changes: fix usage message

PR #615, opened 2024-05-21
nl: -l option unsupported

mknos:
* Usage string incorrectly mentioned "-l num", but this results in "unknown option" error
* GNU and OpenBSD versions support -l; it could be added to this version later if desirable


briandfoy:
changes: undocument unsupported -l switch

briandfoy (approved):

PR #616, opened 2024-05-21
ln: support --version

mknos:
* Move version-info line out of usage() into VERSION_MESSAGE() as done in other scripts; this allows Getopt::Std to handle --version flag
* While here, spell a word correctly in pod and spell a variable differently to avoid conflict with global ($target)

### DIFF
--- a/bin/ln
+++ b/bin/ln
@@ -86,9 +86,9 @@ sub usage {
 }
 
 sub dolink {
-    my ($file, $targ) = @_;
-    return symlink($file, $targ) if $options{'s'};
-    return link($file, $targ);
+    my ($file, $target) = @_;
+    return symlink($file, $target) if $options{'s'};
+    return link($file, $target);
 }
 
 __END__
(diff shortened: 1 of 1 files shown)


bin/ln
```diff
@@ -75,16 +75,20 @@ unless (@ARGV) {
...
-    warn "$Program (Perl bin utils) $VERSION\n";
     warn "usage: $Program [-sf] source_file [target_file | [source_files ...] target_directory]\n";
     exit EX_FAILURE;
 }
 
 sub dolink {
-    my ($file, $target) = @_;
-    return symlink($file, $target) if $options{'s'};
-    return link($file, $target);
+    my ($file, $targ) = @_;
```
briandfoy, line 89: Is there a reason for changing this variable name? I think I prefer `$target`.

briandfoy (changes requested):

bin/ln
mknos: perlcritic identified that a global variable had the same name, so renaming the dolink() param was to avoid confusion

briandfoy (approved):

PR #617, opened 2024-05-23
bc: mod zero

mknos:
* When adding previous commit to avoid divide by zero, mod 0 was missed
* Patch tested under default mode and bignum mode (bc -b)
```
a=123
b=0
a%b
```


briandfoy:
changes: fix regression in modulus 0 operation

PR #618, opened 2024-05-23
ed: edEdit() validation

mknos:
* If I open a file and edit it, the $NeedToSave flag is set
* If I attempt to edit a new file with "e" command, the new file was opened and read into a temporary buffer before validation based on $NeedToSave executes
* I noticed this because I observed a "cannot open file" error when typing a nonexistent filename, instead of a "buffer modified" error
* When editing a new file with "E" command the $NeedToSave flag is not checked, but for "e" command it is
```
%perl ed a.c    # before patch
63
P
*1d
*e aa.cc
aa.cc: No such file or directory
?
*h
cannot open file
*
```

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -666,6 +666,11 @@ sub edEdit {
             edWarn(E_ADDREXT);
             return 0;
         }
+        if ($NeedToSave && $QuestionsMode && !$UserHasBeenWarned) {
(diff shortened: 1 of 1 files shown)


briandfoy:
changes: fix the "need to save" flag for the new file case

PR #621, opened 2024-05-25
cmp: sync usage string

mknos:
* FreeBSD, NetBSD & OpenBSD list usage strings make it clear that flags -l and -s are not compatible
* Code can be simplified slightly by adopting a compatible usage string


briandfoy (approved):

briandfoy:
changes: adjusted the usage example to show hat -l and -s are mutually exclusive

PR #622, opened 2024-05-25
cut: allow -b range of 1

mknos:
* Technically the inclusive byte range of 2-2 is a single byte
* When testing against GNU and OpenBSD versions, the usage of "cut -b 2-2" is allowed and interpreted as "cut -b 2"
* This version raised an error previously but it's more compatible to allow it


briandfoy (approved):

briandfoy:
changes: -b allows a range of 1 now

PR #623, opened 2024-05-26
join: -a or -v

mknos:
* The meanings of options -a and -v are incompatible
* Usage string from standards document indicates -a OR -v [1]; this version prints a compatible usage string
* Follow OpenBSD version and fail early if both -a and -v are given

1. https://pubs.opengroup.org/onlinepubs/009604599/utilities/join.html

### DIFF
--- a/bin/join
+++ b/bin/join
@@ -248,14 +248,19 @@ sub get_field_specs {
 }
 
 sub get_options {
+  my ($aflag, $vflag);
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: don't allow both -a and -v in the some invocation

PR #624, opened 2024-05-28
ed: inferred range commands

mknos:
* Line addresses can be inferred when entering a range with a comma
* An explicit range has two addresses, e.g. "1,2" or "4,$"
* If first address is not given it is always 1
* If first and last address are both not given, last address is $ (last valid line address)
* I discovered this when testing against OpenBSD and GNU ed
* test1: ",n" --> interpret as 1,$n
* test2: ",5n" --> interpret as 1,5n

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -938,11 +938,17 @@ sub edParse {
         $command = '/';
         undef $adrs[0];
     }
+    my $commarange = $fields[1] =~ /,/;
+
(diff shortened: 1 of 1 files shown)


briandfoy:
changes: commands can use inferred line ranges when you omit the first or last value in n,m

briandfoy (approved):

PR #626, opened 2024-05-29
expr: avoid divide by zero

mknos:
* Check for 0 before divide and modulo operations
* Add newlines to die() to prevent line numbers appearing in "syntax error" messages


briandfoy (approved):

briandfoy:
changes: catch a divide-by-zero situation

PR #627, opened 2024-05-29
grep -f 0

mknos:
* It is valid for a pattern file to be named "0" but this was not being ignored by grep
```
%cat > 0
main
puts
^D
%perl grep -nf 0 a.c
4:main(void)
6:	puts("yo");
```


briandfoy (approved):

briandfoy:
changes: allow the search pattern "0" by checking for a defined arg not the arg value

PR #628, opened 2024-05-29
tac: group optional flags in usage string

mknos:
* -B flag can be grouped with -b and -r (all may be provided together)


briandfoy (approved):

briandfoy:
changes: not that -B can be combined with -b and/or -r

PR #631, opened 2024-05-31
cat: add --version

mknos:
* Listing version number in the usage string was a bad idea


briandfoy (approved):

briandfoy:
changes: add --version, adjust usage message

PR #632, opened 2024-05-31
rm: raise error for long options

mknos:
* I noticed that if I provide "long" options to rm, it gets confused and attempts to remove files starting with a dash
* When debugging this, preprocess_options() was incorrectly deciding to save long options into new_args list; call usage() instead because this rm only supports single letter options
```
%perl rm --this    # before patch
rm: cannot remove '-t': No such file or directory
rm: cannot remove '-h': No such file or directory
rm: cannot remove '-i': No such file or directory
rm: cannot remove '-s': No such file or directory
```


briandfoy:
changes: disable "long" options; single letter options only

PR #633, opened 2024-06-02
grep: fix regression for -F search

mknos:
* The following search should match, but did not because of the '#' character: perl grep -F '#include' a.c
* This was because quotemeta() was escaping it as ```\#``` which was then expanded between single quotes
* The $matcher closure can refer to patterns list directly; this avoids having to expand pattern strings within quotes
* This version is easier to follow because duplicate statements are not added to the closure code if many patterns are specified
```
%perl head -n1 a.c
#include <stdio.h>
%perl grep -F '#include' a.c
#include <stdio.h>
%perl grep -Fi '#includE' a.c
#include <stdio.h>
```


briandfoy:
changes: be more careful with quotemeta and patterns

PR #634, opened 2024-06-03
grep: fix -Fc regression

mknos:
* Combining -F and -c flags had the incorrect result of showing matching lines followed by a count
* When debugging this I discovered the opt hash was supposed to be massaged so s-flag is set if c-flag is set
```
%perl grep -Fc '/' aa.c     # before patch
	/* I SHOULD BE INVISIBLE */
1
```

### DIFF
--- a/bin/grep
+++ b/bin/grep
@@ -147,6 +147,7 @@ sub parse_args {
 	@opt{ split //, $nulls } = ('') x length($nulls);
 
 	getopts( $optstring, \%opt ) or usage();
+	$opt{'s'} = 1 if $opt{'c'};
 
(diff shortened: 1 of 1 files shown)


itsnicksia (approved):

itsnicksia:
This fixes an issue triggered on line 400 where line is still printed because only opt `s` is checked:

```
$opt->{'s'} || print $Mult && "$name:",
				$opt->{n} ? "$.:" : "", $_,
				( $opt->{p} || $opt->{P} ) && ( '-' x 20 ) . "\n";

```

briandfoy (approved):

briandfoy:
changes: set -s implicitly when -c is used

PR #635, opened 2024-06-04
grep: fix -Fh

mknos:
* When combining grep -F and -h flags, the filenames were not listed for matches (correct), but they still were not listed when unsetting -h (incorrect)
* test1: "perl grep -F '/' awk ar" --> output should show "$filename:" prefix
* Init of global variable $Mult was missed due to early return in parse_args()
* I also tested that -Fl works as expected
* While here, remove the need for variable $optstring by passing options directly to getopts()

### DIFF
--- a/bin/grep
+++ b/bin/grep
@@ -130,7 +130,7 @@ EOF
 ###################################
 
 sub parse_args {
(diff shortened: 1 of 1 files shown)


briandfoy:
changes: show filenames even without -h set

briandfoy (approved):

PR #636, opened 2024-06-04
tr: pod update

mknos:
* Remove mention of open() failures in BUGS section of manual
* Similar to bin/tee and bin/xargs, change the loop to read STDIN to make it more obvious that tr does not open any files
* On my Linux system, if I redirect stdin from /etc/shadow as a regular user, bash shows a permissions error and sets $? to 1; this takes place before perl is started
* Remove header comment that is duplicated in pod

### DIFF
--- a/bin/tr
+++ b/bin/tr
@@ -11,9 +11,6 @@ License: perl
 
 =cut
 
-
-# tr - translate or delete characters
-
 use strict;
 
 use File::Basename qw(basename);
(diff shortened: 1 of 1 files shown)


bin/tr
```diff
@@ -57,7 +54,7 @@ my $cflag = ($opt{'C'} || $opt{'c'}) ? 'c' : '';
 my $dflag = $opt{'d'} ? 'd' : '';
 my $sflag = $opt{'s'} ? 's' : '';
 eval qq{
-    while (<>) {
+    while (<STDIN>) {
```
briandfoy, line 57: nice catch

briandfoy (approved):

briandfoy:
changes: tr only reads from standard input now

PR #637, opened 2024-06-04
ed: add --version

mknos:
* Simplify getopt by removing -v flag which BSD versions of ed don't support (neither is -v a standard ed flag)
* Remove TODO comment related to adding 'x' command; this command is not supported on OpenBSD (nor FreeBSD from what I see in its manual)
* Remove comment with outdated url
* Sync SYNOPSIS and usage string


briandfoy:
changes: remove -v in favor of --version

briandfoy (approved):

PR #638, opened 2024-06-04
cat: add raw mode

mknos:
* Implement idea from BSD version of cat: bypass input processing if none of options -b -e -n -s -t & -v are provided


briandfoy:
changes: bypass input processing if none of options -b -e -n -s -t & -v are provided

PR #639, opened 2024-06-04
xargs: initial manual

mknos:
Quick'n'dirty pass at documenting the features of this xargs.
The -I option appears to work but I'm not familiar with it.

### DIFF
--- a/bin/xargs
+++ b/bin/xargs
@@ -107,3 +107,53 @@ while (!$eof) {
 =head1 NAME
 
 xargs - construct argument list(s) and execute utility
+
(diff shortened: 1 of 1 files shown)


bin/xargs
```diff
@@ -107,3 +107,53 @@ while (!$eof) {
 =head1 NAME
 
 xargs - construct argument list(s) and execute utility
+
```
briandfoy, line 110: Very nice. Thanks,

briandfoy (approved):

briandfoy:
changes: add pod docs

PR #640, opened 2024-06-05
xargs: retire -l option

mknos:
* Remove non-standard -l alias for -L which is not supported by OpenBSD or NetBSD versions
* The usage string never documented -l so users were theoretically not aware of it
* The code being deleted possibly had a bug where -I option would set $o{l} to 1 but only the value of $o{L} was being used in the main loop

### DIFF
--- a/bin/xargs
+++ b/bin/xargs
@@ -29,7 +29,7 @@ use constant EX_FAILURE => 1;
 my $Program = basename($0);
 
 my %o;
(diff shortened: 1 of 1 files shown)


briandfoy:
changes: remove hidden alias -l for -L since it's not a BSD option

briandfoy:
Ah heck, I merged before approving, but this is approved.

PR #642, opened 2024-06-05
cp: support '--' terminator

mknos:
* '--' option terminator is ignored when executing an external command
* When debugging this I discovered the command list passed to system() strips off the '--' (technically Getopt::Long does)
* Explicit terminator can be added between options and filenames
```
%perl cp -- -a ..   # before patch, on linux
cp: missing destination file operand after '..'
Try 'cp --help' for more information.
```


briandfoy (approved):

briandfoy:
changes: stop option parsing after --

PR #644, opened 2024-06-06
install: remove -M flag

mknos:
* The -M flag is a stub and was included for compatibility; however, version of install from GNU, Solaris[1] and OpenBSD do not support -M
* -M was never mentioned in the usage string or SYNOPSIS but it can be removed entirely
* NetBSD install has -M but it takes a value, so would not be compatible with this stub

1. https://shrubbery.net/solaris9ab/SUNWaman/hman1b/install.1b.html

### DIFF
--- a/bin/install
+++ b/bin/install
@@ -17,7 +17,7 @@ use strict;
 use File::Basename qw(basename dirname);
 use Getopt::Std qw(getopts);
 
-our $VERSION = '1.3';
+our $VERSION = '1.4';
 my $Program = basename($0);
 
 sub VERSION_MESSAGE {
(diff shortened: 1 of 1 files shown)


bin/install
```diff
@@ -17,13 +17,16 @@ use strict;
 use File::Basename qw(basename dirname);
 use Getopt::Std qw(getopts);
 
-my ($VERSION) = '1.3';
+our $VERSION = '1.3';
```
briandfoy, line 20: Update version now that a feature has been removed

briandfoy (changes requested):

bin/install
mknos: Sure, committed.

briandfoy (approved):

PR #646, opened 2024-06-08
od: implement -a option

mknos:
* BSD and GNU versions support -a for printing literal characters, with control character names printed
* GNU version deliberately masks highest (8th) bit per byte, so we follow this
* This is consistent with standards document[1] wording: "named characters from the International Reference Version (IRV) of the ISO/IEC 646:1991 standard. Only the least significant seven bits of each byte shall be used"
* Standards document doesn't mention -a flag, but generally -a is taken to mean the same as "-t a"
* Attempt to document the available options
* When testing this against GNU "od -An -a", whitespace alignment of characters was different but "diff -w" showed no difference for a random test file

1. http://ktiml.mff.cuni.cz/~kucerap/unix/susv4tc2/utilities/od.html

### DIFF
--- a/bin/od
+++ b/bin/od
@@ -23,8 +23,10 @@ use constant EX_FAILURE => 1;
 use constant LINESZ => 16;
 use constant PRINTMAX => 126;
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: add -a option to print control character names

PR #648, opened 2024-06-11
mail: exit command

mknos:
* On OpenBSD the exit command in mail has 3 aliases: x, ex, xit[1]
* I'm not sure why someone would type "xit" instead of "x", but add it here too for completeness

1. http://man.openbsd.org/mail.1


briandfoy (approved):

briandfoy:
changes: support `xit` as a shorter form of `exit`

PR #649, opened 2024-06-11
cp: bundled options

mknos:
* Single-letter options can be grouped; one combo I type sometimes is -iv
 * Tell Getopt::Long to allow this
```
%perl cp -iv ar ar.rar
'ar' -> 'ar.rar'
```


briandfoy (approved):

briandfoy:
changes: support bundled options

PR #650, opened 2024-06-11
words: throw error for too many args

mknos:
* words command takes only one argument; show an error if user accidentally provides 2 arguments instead of ignoring the second
* Zero-args case was already handled


briandfoy (approved):

briandfoy:
changes: die if there are two many arguments

PR #651, opened 2024-06-12
nl: -s separator versus 0

mknos:
* test1: no space between number and line (perl nl -s '' a.c)
* test2: 0 between number and line (perl nl -s0 a.c)
* Patch tested against GNU nl


briandfoy (approved):

briandfoy:
changes: -s only uses default separator when -s is not specified

PR #653, opened 2024-06-14
tail: tweak filename title

mknos:
* Make tail use "==> FILENAME <==\n" format when printing output from multiple files; this conforms with GNU and OpenBSD versions
* This is also consistent with the filename title printed by bin/head
* test: same md5 sum compared to tail(1) on my Linux system: "perl tail -n 50 awk ar arithmetic | md5sum"
* While here, delete dead code in parse_args(); $point is always assigned a default value of -10 at top of function

### DIFF
--- a/bin/tail
+++ b/bin/tail
@@ -125,8 +125,6 @@ sub parse_args()
     }
 
     usage(EX_FAILURE, 'The number cannot be zero') if $point == 0;
-
-    $point = $point ? $point : -10;
     }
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):
Nice. I like the consistency you're adding to PerlPowerTools.

briandfoy:
changes: filename header for multiple file output now matches GNU and BSD, and the head program in PerlPowerTools

PR #654, opened 2024-06-15
tail: retire -h flag

mknos:
* The standard tail command doesn't support -h [1]
* GNU and OpenBSD versions print "unknown option" error
* With this patch, usage string is printed after the error, which matches OpenBSD version
* Simplify usage() so it only takes a single (optional) error message param

1. https://pubs.opengroup.org/onlinepubs/007908799/xcu/tail.html


briandfoy (approved):

briandfoy:
changes: remove the -h option to be consistent with GNU / BSD

PR #655, opened 2024-06-16
ed: 2-address insert command

mknos:
* When testing against GNU and OpenBSD versions, "i" command was with a 2-address prefix was treated as invalid
* Be compatible and treat effective address as 2nd address if 2 were specified
* ed commands "a" and "i" are both handled in edInsert(), so this patch covers both
* test1: "1,2i" --> 2i
* test2: ",2i" --> 2i
* test3: "1,2a" --> 2a
* test4: ",2a" --> 2a

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -749,13 +749,12 @@ sub edInsert {
     my $append = shift;
     my(@tmp_lines);
 
-    if (defined($adrs[1])) {
-        edWarn(E_ADDREXT);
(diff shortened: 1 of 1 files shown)


briandfoy:
changes: if i or a command get two addresses, use the second one as the effective address; GNU/BSD compatibility

briandfoy (approved):

PR #657, opened 2024-06-18
ls: add --version

mknos:
* Now users can inspect installed without having to read source
* Based on a recent patch to bin/cat


briandfoy (approved):

briandfoy:
changes: support --version

PR #659, opened 2024-06-26
ed: command mode versus ctrl+d

mknos:
* Fix the error 'Can't "last" outside a loop block at ed line 189', triggered by ctrl+d (eof) after loading a file in ed
* When adding the input() function the "last" statement was missed in the error handling code
* Consistently clear globals before calling edParse(); prevent saved state from previous command input from being used by subsequent commands
* Follow OpenBSD ed and make ^D equivalent to typing the regular "q" command
* Example case: Open a text file and delete 1st line with "1d" command, then hit ^D
* The first time I hit ^D in command mode, edQuit() will print a warning about the modified buffer (called via edQuitAsk())
* When I hit ^D a second time, edQuit() remembers that a warning was already printed, so the program exits

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -90,8 +90,8 @@ my $Error = undef;              # saved error string for h command
 my $Prompt = undef;             # saved prompt string for -p option
 my $SupressCounts = 0;          # byte counts are printed by default
 my @lines;                      # buffer for file being edited.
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: treat Control-D as the q command so ed exits gracefully

PR #660, opened 2024-06-26
ls: -1 for non-tty

mknos:
* The manual explains that if output is not a tty, ls will infer the -1 flag (i.e. print one entry per line)
* The old filemode check code did not work on my Linux system but this patch does
* I tested this with a pipe into cat, and with a redirect into a file


briandfoy (approved):

briandfoy:
changes: infer -1 switch automatically if the output filehandle is not a tty

PR #661, opened 2024-06-26
od: standard -s option

mknos:
* Standards document outlines the -s flag is intended for 2-byte signed decimal display [1]
* OpenBSD, FreeBSD, GNU and Solaris support this meaning of -s [2]
* The -i flag already follows OpenBSD and prints 2-byte signed decimal, so make -s equivalent to -i
* NB: FreeBSD and GNU treat -i as decimal-of-size-int instead of decimal-of-two-bytes, but -i is not standardised

1. https://pubs.opengroup.org/onlinepubs/009696899/utilities/od.html
2. https://shrubbery.net/solaris9ab/SUNWaman/hman1/od.1.html

### DIFF
--- a/bin/od
+++ b/bin/od
@@ -24,7 +24,7 @@ use constant LINESZ => 16;
 use constant PRINTMAX => 126;
 
 use vars qw/ $opt_A $opt_a $opt_b $opt_c $opt_d $opt_f $opt_i $opt_j $opt_l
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: add -s option and make it the same as -I (two byte signed decimal display)

PR #662, opened 2024-06-27
od: support -B alias for -o

mknos:
* Free/Net/Open BSDs document that "od -B" behaves the same as -o
* GNU version appears to support this as well, but I couldn't find reference to this in its manual


briandfoy (approved):

briandfoy:
changes: add -B as alias for -o

PR #663, opened 2024-06-29
od: add double precision float dump

mknos:
* GNU and OpenBSD versions of od support -e and -F for printing two 8-byte floating point numbers per line
* Write a new formatting function to handle this
* While here, add whitespace for readability of -f flag (regular float dump) and increase the precision digits to 7 to match GNU


briandfoy:
changes: add -e and -f for printing two 8-byte floating point numbers per line

PR #664, opened 2024-07-08
od: add 4-byte hex dump option

mknos:
* GNU and OpenBSD support "od -H" and "od -X" options for 4-byte hex format, so add it here too
* Rename 2-byte hex format function to avoid conflict with perl built-in hex()
* Bump program version

### DIFF
--- a/bin/od
+++ b/bin/od
@@ -24,9 +24,9 @@ use constant LINESZ => 16;
 use constant PRINTMAX => 126;
 
 use vars qw/ $opt_A $opt_a $opt_B $opt_b $opt_c $opt_d $opt_e $opt_F $opt_f
- $opt_i $opt_j $opt_l $opt_N $opt_o $opt_s $opt_v $opt_x /;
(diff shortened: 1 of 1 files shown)


bin/od
```diff
@@ -24,9 +24,9 @@ use constant LINESZ => 16;
 use constant PRINTMAX => 126;
 
 use vars qw/ $opt_A $opt_a $opt_B $opt_b $opt_c $opt_d $opt_e $opt_F $opt_f
- $opt_i $opt_j $opt_l $opt_N $opt_o $opt_s $opt_v $opt_x /;
+ $opt_H $opt_i $opt_j $opt_l $opt_N $opt_o $opt_s $opt_v $opt_X $opt_x /;
 
-our $VERSION = '1.0';
+our $VERSION = '1.1';
```
briandfoy, line 29: We might consider a big bump here since there are new features. But, since the new features don't interfere with current behavior, a minor bump is fine too.

briandfoy (approved):

briandfoy:
changes: add -H and -X options for 4-byte hex output

PR #665, opened 2024-07-09
od: fix data padding

mknos:
* Odd length input was not producing the correct output in all cases; I found this with od -l
* Make a general padding function to avoid duplication in the formatting functions
* Single-byte formatting functions, such as octal1() and char1(), don't require any padding

```
%wc -c a.c
197 a.c
%od -l a.c | tail -n 2   # gnu reference
0000300  1933534308          10
0000305
%perl od -l a.c | tail -n 2   # before patch; no value of 10 in output
00000300 1933534308 
00000305
```

### DIFF
--- a/bin/od
+++ b/bin/od
@@ -285,91 +285,50 @@ sub char7bit {
 }
 
 sub udecimal {
-    if (length($data) & 1) { # pad to 16 bit
-        @arr = unpack 'S*', $data . "\0";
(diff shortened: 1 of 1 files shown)


briandfoy (approved):
very nice refactoring.

briandfoy:
changes: fix odd cases in output padding

PR #666, opened 2024-07-11
od: add 4-byte octal dump

mknos:
* OpenBSD/FreeBSD/NetBSD and GNU versions support big -O for 4-byte octal, and little -o for 2-byte octal
* Patch appears to produce equivalent output to od on my linux system


briandfoy:
changes: add -o and -O for four-byte output

briandfoy (approved):

PR #670, opened 2024-07-18
rm: zero args versus -f

mknos:
* NetBSD, OpenBSD and GNU versions do exit(0) for "rm -f" with no file arguments
* Standards document says -f flag should not modify the exit status for nonexistent operands[1]
* Support this here too
* Make the program more consistent by exiting directly via exit() instead of mixing calls to exit() and $self->exit() (definition of custom exit() can be eliminated)
* Behave more like BSD version by printing usage for bare "rm" with no -f and no args

1. https://pubs.opengroup.org/onlinepubs/009695399/utilities/rm.html

### DIFF
--- a/bin/rm
+++ b/bin/rm
@@ -86,12 +86,13 @@ sub run {
 	$self->error( "$Program: -P ignored\n" ) if $self->is_overwrite;
 
 	unless ( () = $self->files ) {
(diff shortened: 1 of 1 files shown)


briandfoy:
changes: -f with no arguments exits with 0 and shows usage

PR #672, opened 2024-07-19
glob: pod improvement

mknos:
* Remove the section with default values for historical MacOS; people using current Mac would use the suggestions for Unix
* While here, remove the extra empty line in the suggested values for DOS


briandfoy (approved):

briandfoy:
changes: remove docs for use with Mac Classic

PR #675, opened 2024-07-21
split: out-of-order check

mknos:
* The check for the special argument '-' should take place before stat
* The old code was wrong because "split -" produced 2 different results depending on whether I had a directory called "-"
* test1: "perl split -"  == stdin
* test2: "perl split ./-" == file called "-"
* test3: "perl split -- -" == same as test1

### DIFF
--- a/bin/split
+++ b/bin/split
@@ -110,12 +110,12 @@ unless (defined $prefix && length $prefix) {
     $prefix = 'x';
 }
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: allow for a directory named - without confusing it for stdin.

PR #680, opened 2024-07-24
cp: warn on File::Copy::copy() failure

mknos:
* When cp runs in non-unix mode, take advantage of File::Copy::copy() setting $! on failure
* When the destination was a non-existent directory, cp would do exit(1) but did not show the reason for failure


briandfoy (approved):
Yeah, I like that you are cleaning up these long range things.

PR #682, opened 2024-07-25
tee: standard -i flag

mknos:
* Standards document mentions that only SIGINT is ignored for "tee -i" [1]
* Dragonfly/Free/Net/Open- BSD manuals are consistent and list only SIGINT
* Other historical systems documented that -i ignored interrupts but did not explicitly say which ones


briandfoy (approved):

briandfoy:
changes: tee now only ignores SIGINT when -I is specified

PR #683, opened 2024-07-26
tee: remove nonstandard -n option

mknos:
* There is no need to support -n, which is not provided by the BSD or GNU versions (nor AIX [1], Solaris)
* Make some changes to prepare for enabling strict mode, e.g. add missing declaration for $status variable

1. https://www.ibm.com/docs/en/aix/7.2?topic=t-tee-command

### DIFF
--- a/bin/tee
+++ b/bin/tee
@@ -17,28 +17,18 @@ License: perl
 # Tom Christiansen <tchrist@convex.com>
 # 6 June 91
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):
And, on macOS the -n is a completely different thing for nonseekable files.

briandfoy:
changes: remove non-standard -n, which suppressed standard output.

PR #684, opened 2024-07-27
tee: add strict

mknos:
* Correct left-over reference to variable $unbuffer
* Remove custom SIGPIPE code which doesn't exist in the BSD version (the only reference to signals is SIGINT being ignored)
* Standards document mentions nothing about SIGPIPE in Asynchronous Events section[1]
* Output files table (%fh) is now keyed based on file descriptor and contains filehandle and the saved filename from ARGV
* Explicitly check for directory before opening a file
* test1: "ifconfig | perl tee out.1" --> out.1 created
* test2: "ifconfig | perl tee -a out.1 out.2" --> existing out.1 appended, out.2 created

### DIFF
--- a/bin/tee
+++ b/bin/tee
@@ -17,44 +17,57 @@ License: perl
 # Tom Christiansen <tchrist@convex.com>
 # 6 June 91
 
+use strict;
+
 use Getopt::Std qw(getopts);
+use IO::File;
 
 my %opt;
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
I think the SIGPIPE thing was a joke.

PR #687, opened 2024-07-29
sort: documentation fix

mknos:
* Usage string incorrectly advertised that the first file argument is required; all file arguments are optional
* Document that the program will read from stdin by default
* Remove second-person language from manual


briandfoy (approved):

briandfoy:
changes: adjust docs to correctly denote that filename args are optional

PR #688, opened 2024-07-30
sort: reject directory arguments

mknos:
* If any file argument is a directory, sort should print an error and exit
* I found this when testing against GNU and OpenBSD versions


briandfoy (approved):

briandfoy:
changes: die if an argument is a directory

PR #689, opened 2024-07-30
paste: out-of-order check

mknos:
* As with PR675, check for '-' argument needs to occur first
* The argument '-' doesn't have a special meaning for all commands


briandfoy (approved):

briandfoy:
changes: don't check that argument - is a literal filename.

PR #690, opened 2024-07-31
grep: out-of-order check

mknos:
* grep supports '-' argument for stdin
* On L180 '-' is added to ARGV if it is empty
* The directory check in matchfile() was happening too early, so move the '-' check to top of loop
* test1: "mkdir ./- && echo hi | perl grep hi" --> previously the directory check would get in the way and terminate grep
* test2: "ifconfig | perl grep '^[a-z]'" --> ifconfig lines starting with an interface name

### DIFF
--- a/bin/grep
+++ b/bin/grep
@@ -290,7 +290,11 @@ sub matchfile {
 FILE: while ( defined( $file = shift(@_) ) ) {
 		my $compressed = 0;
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: improve handling of - to mean standard input (check was too early)

PR #692, opened 2024-08-01
fold: add getopts()

mknos:
* Reduce the amount of custom code for pre-posix argument handling of -NUM
* Pre-process ARGV by converting -123 to -w123, then hand off to getopts()
* Prefix usage string with "usage:"
* Sync usage string with pod document; no need to encourage people to use the old -NUM form for width
* test1: "echo `perl cat cat` | perl fold -40" --> process stdin with -w 40
* test2: "perl fold -s -w 72 moo" --> regular -w form with file input instead of stdin

### DIFF
--- a/bin/fold
+++ b/bin/fold
@@ -28,59 +28,34 @@ use strict;
 use locale;             # for what a space is.
 
 use File::Basename qw(basename);
+use Getopt::Std qw(getopts);
 
 use constant EX_SUCCESS => 0;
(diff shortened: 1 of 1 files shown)


briandfoy (approved):
All of these programs your catching out on this make me want to find what when GetOpts was part of perl core. I think Getopt::Std was in the first release of Perl 5 though.

briandfoy:
changes: use getopts to handle argument processing

PR #693, opened 2024-08-01
true/false: ignore options

mknos:
* Standards document says that true does not handle options [1]
* No reference to options in the solaris manual either [2]
* I tested 4 versions of true and only the GNU version handles options
* test1: /bin/true --version # gnu version on linux
* test2: builtin true --version # bash builtin on linux
* test3: /usr/bin/true --version # openbsd version
* test4: builtin true --version # ksh builtin on openbsd
* The same is true for false

1. https://pubs.opengroup.org/onlinepubs/009695399/utilities/true.html
2. https://shrubbery.net/solaris9ab/SUNWaman/hman1/true.1.html

### DIFF
--- a/bin/false
+++ b/bin/false
@@ -11,36 +11,7 @@ License: perl
 
 =cut
 
-
-use strict;
-
-my ($VERSION) = '1.3';
-
-my $STATUS = 1;
-
-if (@ARGV) {
-    if ($ARGV [0] eq '--version') {
-        $0 =~ s{.*/}{};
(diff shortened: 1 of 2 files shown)


briandfoy (approved):
I once asked an internal to write me a login shell that would simply output a message saying that the account is restricted and then exit. He delivered three or four screens of code after a month of trying. I then showed him something like this:

```
printf( stdout, .... );
exit(1);
```

PR #695, opened 2024-08-02
fmt: validate width

mknos:
* I observed a suspicious warning when providing negative number as input...
```
%perl fmt -w-1000000 a.c > /dev/null
Negative repeat count does nothing at fmt line 34.
```
* Value for -w must be an int; follow OpenBSD fmt and raise error for 0 and negative
* test1: "perl fmt -w 1.1  a.c" --> already covered by Getopt::Long
* test2: "perl fmt -w''  a.c" --> same as test1
* test3: "perl fmt -w0  a.c" --> new validation
* test4: "perl fmt -w-100  a.c" --> same as test3

### DIFF
--- a/bin/fmt
+++ b/bin/fmt
@@ -30,6 +30,10 @@ Getopt::Long::config('bundling');
 GetOptions(
     'w=i' => \$MAX_WIDTH,
 ) or usage();
+if ($MAX_WIDTH <= 0) {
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: max width must be greater than zero

PR #696, opened 2024-08-03
fmt: exit(1) for directory

mknos:
* Follow pattern from OpenBSD fmt: print warning and guarantee non-zero exit code if a file arguments was a directory (e.g. "perl fmt .. .. text.txt")
* Sync SYNOPSIS with usage string; the -w option is preferred over the traditional -NUM option
* Document that the standard input is read by default
* Add a pod section for EXIT STATUS


briandfoy:
changes: exit with an error if an argument is a directory rather than skipping it

PR #699, opened 2024-08-05
patch: delete dead -x option

mknos:
* The -x (aka --debug) option exists in the GNU version of patch, but the perl version never did anything with -x
* Mentioning -x in the pod manual makes the useful information more difficult to find
* This option is not covered by the standard so removing it is not a compatibility issue [1]
* Debugging the perl version can be as simple as setting a breakpoint in Patch::apply() or Patch::Ed::apply()


briandfoy (approved):

briandfoy:
changes: remove -x option (debugging) that wasn't doing anything

Issue #404, opened 2024-01-09
RT #98905: factor and primes are limited to 2^32


briandfoy:
From DANAJ in [#98905 on rt.cpan.org](https://rt.cpan.org/Ticket/Display.html?id=98905):

Both factor and primes check that the inputs are 10 or fewer digits and less than 2^32.  It is unclear to me whether this is:

  1) intentional behavior that should always be maintained for compatibility.

or

  2) sanity checking the range, and since 64-bit Perl didn't come until after these were written, 32-bit was hard coded.

These are not hard to extend, albeit they're not super speedy for larger numbers, but Ben Tilly wrote some pretty nice code so not nearly as bad as one might expect.  I've been playing with alternate primes implementations -- it can be easily sped up 2x for large bases, and I'm sure more than that if we accept lots of turmoil.

For extra credit, we could even use Math::BigInt, but doing that for all inputs while maintaining decent speed might give too much bloat (at least for my coding it would).

briandfoy:
Hey @mknos, is this one something that you are interested in?

pjacklam:
I played with the `factor` program, and it puzzled me why it became so slow then the input was large. I didn't look closely into the implementation, because I knew I had a `factor` program that was much faster.

First I modified the original program to allow for larger inputs. I changed the test into the following, which allows input values up to 2¹¹² on systems built with quadmath and 2⁶⁴-1 on systems without quadmath, but with for 64 bit integers.

```
   unless ($n =~ /^\+?\d+$/) {
        warn "$0: `$n:' is not a valid positive integer\n";
        return;
    }

    if ($n % 2 == 0 && ($n - 1) % 2 == 0) {
        warn "$0: `$n:' is too large\n";
        return;
    }
```
Then I improved my program and ran it against the current `factor` program. Below are the results. The execution times are in seconds. My program is about 4 times as large as the current `factor` program, which I assume explains why my program is slightly slower for the smallest input values. Note that I didn't test the original `factor` with the largest inputs. It would have taken many hours. I'll update this entry if I get the data.

|  n | largest n-digit prime | factor (new) | factor (orig) |
|  - | - | - | - |
|  1 | 7 | 0.11 | 0.03 |
|  2 | 97 | 0.11 | 0.03 |
|  3 | 997 | 0.11 | 0.03 |
|  4 | 9973 | 0.11 | 0.03 |
|  5 | 99991 | 0.11 | 0.03 |
[...]

PR #700, opened 2024-08-05
factor: increase range and speed up code

pjacklam:
- Fix CPAN RT #98905 / Issue #404 and make factor to handle as large integers possible without using bignum (or similar).

- Use a different algorithm (wheel factorisation of order 6), which is much faster for larger integers.


briandfoy (approved):

briandfoy:
Congratulations on your first PR :)

briandfoy:
changes: completely replace factor with a program use the wheel algorithm to make it speedy (#404), from Peter John Acklam

PR #701, opened 2024-08-06
patch: -o filename versus directory

mknos:
* The "-o file" option is a standard feature of patch
* When I tested -o with a directory argument the directory was replaced with the patched file
* After the initial panic that patch had destroyed the directory, I discovered that it only renamed it to thing.orig
* A directory is not a valid argument for -o; raise an error instead of processing the patch
* This was found when testing against GNU patch


briandfoy (approved):

briandfoy (approved):

briandfoy:
changes: -o complains if its argument is a directory

PR #702, opened 2024-08-06
patch: accept --dry-run option

mknos:
* OpenBSD and FreeBSD versions accept -C, --check and --dry-run; GNU version only accepts --dry-run
* This version already accepted -C and --check


briandfoy (approved):

briandfoy:
changes: --dry-run is an alias for -C or --check

PR #704, opened 2024-08-07
tee: detect write() failure

mknos:
* Follow wording from standards document that write() errors should cause the current file to stop being written to; however, remaining files and stdout should continue
* Switching to syswrite() guarantees the output is unbuffered as expected, and makes detecting failure simpler


briandfoy (approved):

briandfoy:
changes: catch situations where tee cannot write to the destination

PR #705, opened 2024-08-07
bc: remove signal handling

mknos:
* Standard bc does not require any signal handling [1]
* Different versions of bc treat SIGINT differently
* The previous code resulted in bc exiting on ^C on my Linux system, so there's effectively no difference with this patch

1. https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/utilities/bc.html

### DIFF
--- a/bin/bc
+++ b/bin/bc
@@ -2739,16 +2739,6 @@ sub floor_idx {
   return -1;
 }
 
-# catch signals
-sub catch
-{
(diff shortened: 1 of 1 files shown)


briandfoy (approved):
I was thinking that maybe we could catch SIGINT to exit cleanly, but, then, there's nothing we have to clean up and the result would be the same.

PR #706, opened 2024-08-08
patch: simplify --check

mknos:
* "patch --check" can be implemented in less code using File::Temp to provide a garbage filehandle which is automatically cleaned up later
* I tested this by repeatedly running "perl patch -C original.c my.diff"; original.c is not modified so each instance is successful
* For --check mode there is no need for o_fh and d_fh to be a different filehandle because data written to it is discarded anyway

### DIFF
--- a/bin/patch
+++ b/bin/patch
@@ -18,6 +18,8 @@ License: perl
 
 use strict;
 
+use File::Temp qw();
+
 my $VERSION = '0.28';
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: use File::Temp to get the /dev/null filehandle

PR #707, opened 2024-08-08
units: fail if custom unittab file doesn't open

mknos:
* I found this when testing against the OpenBSD version
* The old code unhelpfully bypassed opening the file specified by -f FILE option if it did not exist
* Make read_unittab() handle 3 cases...
* case1: implicit unittab ($file is undefined; read definitions in __DATA__)
* case2: $file is a directory (ignore it to match OpenBSD; read __DATA__)
* case3: otherwise... open the file and terminate the program if it couldn't open
* While here add defined() checks after prompting for input; this silences warnings seen when terminating program with ^D

### DIFF
--- a/bin/units
+++ b/bin/units
@@ -94,13 +94,13 @@ sub run {
     while (1) {
       print "You have: ";
       my $have = <>;
-      exit 0 unless $have =~ /\S/;
+      exit 0 unless defined($have) && $have =~ /\S/;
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: fail for a custom units file only if one was specified but does not exist

PR #708, opened 2024-08-08
grep: -q and -s options reversed

mknos:
* This grep has the standard meanings of options -s and -q backwards [1]
* -s is to suppress some errors
* -q is to make grep not show matches, and perform a quick search of one match per file
* test1: "perl grep -c yo grep" --> count the matches of "yo"
* test2: "perl grep -cv yo grep"-->  count the lines not matching "yo"
* test3: "perl grep -q perl  grep" --> quiet search for "perl"; exit status 0 indicates Match
* test4: "perl grep -q pascal grep" --> quiet search for "pascal"; exit status 1 indicates NoMatch
* test5: "perl grep -s greg greg" --> silent search for "greg" but the file doesn't exist; exit status 2 indicates Error
* Correct POD manual and usage string

1. https://pubs.opengroup.org/onlinepubs/009604299/utilities/grep.html

### DIFF
--- a/bin/grep
+++ b/bin/grep
@@ -97,13 +97,13 @@ sub usage {
 usage: $Me [-incCwsxvhlF1HurtpaqT] [-e pattern]
         [-f pattern-file] [-P sep] [pattern] [file...]
 
-Standard grep options:
+Options:
 	-i   case insensitive
 	-n   number lines
 	-c   give count of lines matching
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: -s and -q were swapped in previous versions. -s is for suppressing errors, -q is for not showing matches

PR #709, opened 2024-08-09
grep: filename prefix correction

mknos:
* grep output is expected to show the filename if grep is searching many files, or in a recursive search (grep -r)
* The -h flag disables the filename prefix
* $Mult flag controls showing the filename, and was being set to 1 sometimes when it shouldn't be
* test1: "perl grep perl awk" --> show matches for 1 file (no filename prefix, single file argument)
* test2: "perl grep  perl awk ar" --> matches for 2 files (filenames included)
* test3: "perl grep -h perl awk ar" --> matches for 2 files (filenames disabled by -h)
* test4: "perl grep -hr perl ." --> recursive search (filenames disabled)
* test5: "echo elmo | perl grep elm" --> show matches for stdin (no filename prefix, no file arguments)
* test6: "perl grep -l awk a*" --> show filenames matching awk
* test7: "perl grep -r package ." --> recursive search (filenames included)

### DIFF
--- a/bin/grep
+++ b/bin/grep
@@ -144,7 +144,6 @@ sub parse_args {
 	@opt{ split //, $nulls } = ('') x length($nulls);
 
 	getopts('incCwsxvhe:f:l1HurtpP:aqTF', \%opt) or usage();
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: show filename if matching multiple files

PR #710, opened 2024-08-10
grep: shortcut exit path for -q

mknos:
* The -q flag in grep is concerned if a match was found in any file
* The program can successfully terminate after the first match without searching subsequent files in argument list
* GNU grep behaves in this way; strace on my Linux system confirms that only the 1st file in argument list is opened when that file contains a match
```
%strace  grep -q perl ar awk 2> trace # gnu grep
%grep open trace
...
openat(AT_FDCWD, "ar", _RDONLY|O_NOCTTY|O_LARGEFILE) = 3
^^^ final openat() in process
```

### DIFF
--- a/bin/grep
+++ b/bin/grep
@@ -403,7 +403,10 @@ FILE: while ( defined( $file = shift(@_) ) ) {
 			}
 		}
 	continue {
-		print $Mult && "$name:", $total, "\n" if $opt->{c};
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: stop looking after first match with -q.

PR #711, opened 2024-08-12
cmp: sysread failure guard

mknos:
* Input loop terminates if sysread on file1 fails
* If sysread failed on file2, $read_in2 would be undef, then $checklength would be undef
* It is not correct to compare the two input buffers in this case; $buffer2 possibly wouldn't have been modified since the previous loop iteration
* Skip the buffer comparison if $checklength is zero (treat sysread failure in the same way because there are no bytes to compare)
* In this case it would be correct for cmp to report "EOF on file2" because the read length for file1 is greater

### DIFF
--- a/bin/cmp
+++ b/bin/cmp
@@ -168,12 +168,13 @@ if ($skip2) {
 }
 
 READ: while (defined ($read_in1 = sysread $fh1, $buffer1, $chunk_size)) {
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: stop if we can't read from the first file

PR #713, opened 2024-08-12
ar: failed read() error

mknos:
* After opening an archive, print $! if it wasn't possible to read the file
* Now it's easier to tell between i/o errors and files which don't contain the magic number data
* Conditionally declare $attr in printList() which is only needed for verbose mode
* test1: "ar t ." --> directory error
* test2: "perl ar t /bin/sh" --> file format error
* test3: "perl ar t test.a" --> short listing of archive contents
* test4: "perl ar tv test.a" --> detailed listing of archive contents

### DIFF
--- a/bin/ar
+++ b/bin/ar
@@ -200,8 +200,8 @@ sub printMember {
 # writes a directory-style listing for the specified archive member
 sub printList {
     my ($name, $pAr, $verbose) = @_;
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: show system error when we can't read one of the files

PR #714, opened 2024-08-12
awk: auto-clean /tmp

mknos:
* Switch from calling tempfile() to File::Temp->new() to allow the input and output files to be unlinked automatically when the program terminates
* This prevents remnant files in /tmp if die() happens

### DIFF
--- a/bin/awk
+++ b/bin/awk
@@ -17,7 +17,7 @@ License: perl
 
 use strict;
 
-use File::Temp qw(tempfile);
+use File::Temp qw();
 
 my(
     $program,		# the code to eval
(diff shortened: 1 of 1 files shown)


bin/awk
```diff
@@ -44,8 +44,10 @@ usage() unless @ARGV;
 open(SAVE_OUT, '>&', STDOUT) || die "can't save stdout: $!";
 die unless defined fileno SAVE_OUT;
 
-my ($tmpin_fh, $tmpin) = tempfile();
-my ($tmpout_fh, $tmpout) = tempfile();
+my $tmpin_fh = File::Temp->new();
```
briandfoy, line 47: I think you need to add `UNLINK => 1` to the constructor to get the auto-cleanup. This is the same as adding that to the call to `tempfile()`. Is there something else going on there?

briandfoy (changes requested):

bin/awk
mknos: The File::Temp manual mentions UNLINK is a default flag for new().
```
        by default the object is constructed as if "tempfile" was called
        without options, but with the additional behaviour that the
        temporary file is removed by the object destructor if UNLINK is set
        to true (the default).
```
If I added an explicit UNLINK to the call, I should probably do it for the other scripts: bin/mail, bin/patch.
When adding a temporary die() statement after the modified code, the files do not appear in /tmp, so to me it proves the auto-unlink.
But for tempfile() the default is UNLINK=0, so the code would need to handle it separately.

briandfoy (approved):

briandfoy:
changes: tempfiles are now removed at the end of the run

PR #715, opened 2024-08-12
id: gid 0 is valid

mknos:
* Previously commit 92678daec769369d29fef632fc9e412b511698b1 was added to allow resolving "root" to uid 0
* When testing -g option a similar issue was discovered (gid of 0 was being treated as a failure of getpwnam)
* This was discovered while testing against id on my Linux system

```
%grep '^root' /etc/group 
root:x:0:
%/usr/bin/id -g root
0
%perl id -g root # before patch
1000
```


briandfoy (approved):

briandfoy:
changes: allow gid to be 0

PR #716, opened 2024-08-13
dirname: support -- terminator

mknos:
* Standards document has an example where "--" option terminator is used.. $(dirname -- file) [1]
* dirname is not a bash builtin on my Linux system
* GNU and OpenBSD dirname support "--" so add it here too
* GNU version supports multiple path arguments, but that does not follow the standard (this version already raised an error for multiple arguments)
* test1: "perl dirname -- $(pwd)/A" --> single file argument after -- is valid
* test2: "perl dirname a b" --> too many args
* test3: "perl dirname --" --> not enough args 
* test4: "perl dirname -x -- -A" --> bad option error

1. https://pubs.opengroup.org/onlinepubs/9699919799/utilities/dirname.html

### DIFF
--- a/bin/dirname
+++ b/bin/dirname
@@ -13,20 +13,29 @@ License: perl
 
 
 use strict;
-use File::Basename;
 
-my ($VERSION) = '1.2';
+use File::Basename qw(basename dirname);
+use Getopt::Std qw(getopts);
 
-unless (@ARGV == 1) {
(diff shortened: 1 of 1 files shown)


briandfoy (changes requested):
Since you've completely replaced the program, you should get credit in the metadata

briandfoy (approved):

PR #718, opened 2024-08-14
patch: chdir happens too late

mknos:
* When testing against GNU and OpenBSD patch commands I found that -d DIR in this version is interpreted after reading a patch file
* The standard says -d is interpreted "before processing" [1]
* Follow the other versions of patch and do chdir() as early as possible:
A) Failure to chdir() will terminate the program before any input is read
B) If -d option is used, files in the argument list are relative to that directory not to the originating directory

* test1: "perl patch -d ... a.c" --> fail (no such dir)
* test2: "patch -d .. a.c a.diff" --> patch ../a.c with ../a.diff

### DIFF
--- a/bin/patch
+++ b/bin/patch
@@ -84,6 +84,11 @@ $patchfile = '-' unless defined $patchfile;
 
 my $patch = Patch->new(@options);
 
+if (defined $patch->{'directory'}) {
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: -d DIR now happens before processing, as it should

PR #719, opened 2024-08-14
patch: reject negative --fuzz value

mknos:
* Fail as early as possible if a negative number was given for --fuzz option
* Getopt::Long already validates --fuzz argument as a number, but  a negative value was previously ignored (the default of 2 would be used)
* Found when testing against GNU patch


briandfoy (approved):

briandfoy:
changes: --fuzz argument must be a positive number

PR #720, opened 2024-08-14
random: raise error for bad option

mknos:
* Comparing the NetBSD version of the random-lines game, unknown options should result in usage string being printed
* Usage was already printed for too-many-arguments case
* Create a regular usage() function as done in other scripts

```
%ifconfig | perl random -r 6 # test1: -r
        RX packets 0  bytes 0 (0.0 B)

        inet6 ::1  prefixlen 128  scopeid 0x10<host>
        RX packets 2057  bytes 452554 (441.9 KiB)
%perl random -e 100 # test2: -e
%echo $?
35
%perl random -x # test3: bad option
Unknown option: x
usage: random [-er] [denominator]
%perl random 123 456 # test4: too many args
usage: random [-er] [denominator]
```

### DIFF
--- a/bin/random
+++ b/bin/random
@@ -15,45 +15,15 @@ License: perl
 use strict;
 use Getopt::Std;
 
-my ($VERSION) = '1.2';
+my ($VERSION) = '1.3';
 
 my $warnings = 0;
-
-END {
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: die for bad options

PR #721, opened 2024-08-15
patch: allow --output to stdout

mknos:
* Allow '-' to be interpreted as standard out for -o option (aka --output); this works for Linux and OpenBSD versions
* GNU diff writes notifications to stderr so they aren't jumbled with patch output; update note() to follow this
* When writing to standard out, don't check if there's a file called '-' in current directory because it is not correct to make a backup by renaming it

### DIFF
--- a/bin/patch
+++ b/bin/patch
@@ -395,6 +395,7 @@ sub bless {
         }
     }
 
+    my $to_stdout = 0;
     my ($in, $out);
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: use - to send output to standard output

PR #723, opened 2024-08-16
basename: support -- option terminator 

mknos:
* Similar to what was committed for bin/dirname, '--' should be interpreted as the end of options (currently it is taken literally)
* This was found when testing against OpenBSD version
* test1: "perl basename --" --> error, no argument
* test2: "perl basename -x" --> error, bad option
* test3: "perl basename -- -x" --> basename of "-x"
* While here, one call to usage() in bin/dirname was missed

### DIFF
--- a/bin/basename
+++ b/bin/basename
@@ -13,19 +13,15 @@ License: perl
 
 
 use strict;
-use File::Basename;
 
-my ($VERSION) = '1.3';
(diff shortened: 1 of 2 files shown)


briandfoy (approved):

briandfoy:
changes: accept -- as end of options

PR #724, opened 2024-08-19
basename: suffix argument did nothing

mknos:
* The pod manual mentioned that the optional suffix could be a normal string or a regex
* Follow the standards document and implement string match, being careful to match only at the end of the string, and not to match the entire string (after prefix is stripped) [1]
* Remove reference to regex suffix in pod
* Bump version
```
%perl basename `pwd`/a.c  # test1: no suffix
%perl basename `pwd`/a.c eyy.si  # test2: no match for suffix
%perl basename `pwd`/a.c a.c  # test3: match consumes all of remaining string so it is ignored
%perl basename `pwd`/a.c .c   # test4: file extension .c is removed
```
1. https://pubs.opengroup.org/onlinepubs/007904975/utilities/basename.html "If the suffix operand is present, is not identical to the characters remaining in string..."

### DIFF
--- a/bin/basename
+++ b/bin/basename
@@ -18,15 +18,13 @@ use File::Basename qw(basename fileparse);
 use Getopt::Std qw(getopts);
 
 my $Program = basename($0);
-our $VERSION = '1.4';
+our $VERSION = '1.5';
 
 getopts('') or usage();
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: actually support the optional suffix argument

PR #727, opened 2024-08-28
ar: -t lists files not in archive

mknos:
* The -t option lists all archive members by default, otherwise only the files specified as arguments
* printList() was incorrectly printing a name that was not found in the archive
* binutils ar prints a warning in this case (printMember() follows this but printList() did not)
```
%perl ar -t test.a # test archive contains 2 files words
wordlist.txt.old
%perl ar -t test.a words words2 # output indicates words2 exists in archive
words
words2
%ar -t test.a words words2 # binutils comparison
words
no entry words2 in archive
```

### DIFF
--- a/bin/ar
+++ b/bin/ar
@@ -200,8 +200,13 @@ sub printMember {
 # writes a directory-style listing for the specified archive member
 sub printList {
     my ($name, $pAr, $verbose) = @_;
+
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: -t correctly reports missing members now

PR #729, opened 2024-08-30
ar: best-effort extract & delete

mknos:
* If extracting or deleting multiple archive members, process all arguments even if some members are not found
* This patch follows how binutils version works
```
%ar -t test2.a # list archive members
words
wordlist.txt.ol
%ar -xv test2.a not1 not2 words # extract 3 files (1 is extracted)
no entry not1 in archive
no entry not2 in archive
x - words
%ar -dv test2.a test test2 words # delete 3 files (1 is deleted)
No member named `test'
No member named `test2'
d - words
```

### DIFF
--- a/bin/ar
+++ b/bin/ar
@@ -83,7 +83,7 @@ for my $file (@ARGV) {
 			print "d - $name\n" if $opt_v;
 			}
 		else {
-			die "$0: $name: not found in archive\n";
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: extract/delete all arguments even if some members are not found

PR #731, opened 2024-09-04
pwd: don't raise error for -- terminator

mknos:
* This version of pwd already acted like the BSD version because it would raise an error if any non-option arguments are provided
* When testing against OpenBSD, the usage ```pwd -L --``` is valid; this version raised an error because '--' was interpreted as an argument instead of option terminator
* getopts() handles -- internally, so switch over to it as done in other scripts

### DIFF
--- a/bin/pwd
+++ b/bin/pwd
@@ -13,30 +13,16 @@ License: perl
 
 
 use strict;
-use Cwd;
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: allow --, even though there are no arguments

PR #733, opened 2024-09-07
addbib: support ed editor

mknos:
* BSD version of addbib allows vi, ex, edit and ed [1] [2]
* This version supported emacs (keep it) but not ed

1. https://github.com/AaronJackson/2.11BSD/blob/main/usr.bin/refer/addbib.c#L215
2. https://man.freebsd.org/cgi/man.cgi?query=addbib&apropos=0&sektion=0&manpath=FreeBSD+14.1-RELEASE+and+Ports&arch=default&format=html

### DIFF
--- a/bin/addbib
+++ b/bin/addbib
@@ -104,7 +104,7 @@ while (1) {
 } continue {
 	print "Continue? (y) ";
 	$_ = <>;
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: add ed to the supported editors

PR #735, opened 2024-09-12
col: retire options -E/-e

mknos:
* By default both types of escapes are supported
* Standard col provides no flag for disabling particular escapes [1], so it's not necessary here either
* BSD versions of col support both types of escapes so follow that [2] [3] [4]
* BSD versions support a greater number of escapes than this version, but more escapes could be added here later
* Sync usage string and POD document now that -E and -e are gone; extension options -s and -t remain
* Bump version number

1. https://pubs.opengroup.org/onlinepubs/7990989799/xcu/col.html
2. https://man.freebsd.org/cgi/man.cgi?query=col&apropos=0&sektion=0&manpath=FreeBSD+14.1-RELEASE+and+Ports&arch=default&format=html
3. https://man.netbsd.org/col.1
4. http://man.openbsd.org/col

### DIFF
--- a/bin/col
+++ b/bin/col
@@ -21,12 +21,11 @@ use constant EX_SUCCESS => 0;
 use constant EX_FAILURE => 1;
 
 my $Program = basename($0);
-our $VERSION = '1.1';
+our $VERSION = '1.2';
 
-use vars qw($opt_b $opt_f $opt_p $opt_x $opt_l);
-use vars qw($opt_e $opt_E $opt_s $opt_t);
+use vars qw($opt_b $opt_f $opt_l $opt_p $opt_s $opt_t $opt_x);
 
-getopts('bfpxl:eEst') or usage();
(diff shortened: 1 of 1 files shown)


briandfoy (changes requested):
I'm wondering if we should accept the -e and -E as legacy options and simply do nothing with them. If someone is already using them, they don't get an error. Thoughts?

mknos:
If other versions of col had -e/-E options it might be worth keeping these as compat stubs, but I can't find any versions that have it. On Linux I have "col from util-linux 2.36.1", which doesn't have it. It's not documented on AIX, Solaris or Plan9 either, so unix users would not be familiar with these options. I prefer keeping the patch as is.

https://www.ibm.com/docs/fi/aix/7.1?topic=c-col-command
https://shrubbery.net/solaris9ab/SUNWaman/hman1/col.1.html
https://9fans.github.io/plan9port/man/man1/col.html

PR #736, opened 2024-09-14
col: add -h option

mknos:
* GNU and BSD versions of col support -h which is the same as the default operation of col
* In those versions -h and -x given together negate each other, so the final one wins
* getopts() doesn't disclose the order of options, so document that -x takes precedence over -h if both are given
* Sync usage() function
* test1: "perl col < col | md5sum" --> normal output
* test2: "perl col -h < col | md5sum" --> same output as test1
* test3: "perl col -x < col | md5sum" --> multiple spaces output
* test4: "perl col -xh < col | md5sum" --> honour -x despite -h coming last; same output as test3

### DIFF
--- a/bin/col
+++ b/bin/col
@@ -23,9 +23,9 @@ use constant EX_FAILURE => 1;
 my $Program = basename($0);
 our $VERSION = '1.2';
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: add -h for compatibility, although that's what we already do

PR #737, opened 2024-09-17
col: dead buffering code

mknos:
* Partial file buffering code was removed in commit 5da850a4719b91919d42c3cbbe993cc5a33493de
* Argument to option -l was being validated as a number, but $opt_l was effectively unused
* Remove validation code but it's not harmful keeping -l option as a stub since it appears in other versions (e.g. GNU and FreeBSD)
* Update description of -l in POD

### DIFF
--- a/bin/col
+++ b/bin/col
@@ -23,19 +23,9 @@ use constant EX_FAILURE => 1;
 my $Program = basename($0);
 our $VERSION = '1.2';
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: -l is now a no-op, but preserved for legacy operation.

PR #738, opened 2024-09-18
col: add --version

mknos:
* --version flag is supported in other scripts so add it here too
* style: $_  can be omitted for substitution


briandfoy (approved):

briandfoy:
changes: add --version

briandfoy (approved):

PR #739, opened 2024-09-20
grep: retire study()

mknos:
* Current perl treats study() call as no-op so delete it [1]
* grep was the only caller of study(); no other scripts need updating
* While here, convert a "use" statement within a condition to a "require" (found by perlcritic)
* Regression test: "perl grep -Hn include a.c b.c" --> still highlights the matching part of each line on my Linux system

1. https://perldoc.perl.org/functions/study


briandfoy (approved):
Heh - https://stackoverflow.com/questions/8383527/which-situations-benefit-from-perls-study

PR #743, opened 2024-09-24
grep: common meaning for -H

mknos:
* BSD and GNU greps support -H for always showing filename headers, and -h for never showing filename headers
* This version had -h with the same meaning, but no option for always printing filenames
* The -H option was being used for enabling terminal highlighting of matching sections per line
* There's no standard way of toggling terminal highlighting for grep, so re-letter highlighting to -g and make -H the opposite of -h
* Update pod and usage string
* test1: "perl grep -h include a.c" --> same as default; no filename header is printed because only 1 argument
* test2: "perl grep -h include a.c b.c" --> suppress filenames and only print matching text for the 2 files
* test3: "perl grep -H include a.c" --> force grep to show filename header for single argument
* test4: "perl grep -H include a.c b.c" --> same as default; filename headers are printed because >1 file argument

### DIFF
--- a/bin/grep
+++ b/bin/grep
@@ -53,7 +53,7 @@ use File::Basename qw(basename);
 use File::Spec;
 use Getopt::Std;
 
-our $VERSION = '1.003';
+our $VERSION = '1.004';
 
 $| = 1;                   # autoflush output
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: -H moved to always show filenames, and highlighting is now -g

briandfoy:
Good catch. I use -H a lot.

PR #744, opened 2024-09-24
grep: silence stdin warning

mknos:
* Make this grep follow the BSD version better
* A warning is already printed under trace mode (-T option) for each file argument processed, but there is no need to warn by default
* GNU grep doesn't warn for this test case either
* Fix a nit where the trace message is supposed to use the variable $name
* While here, consistently append "\n" to error messages 
```
%ifconfig | perl grep -T flags # after patch ...
grep: checking <STDIN>
eth0: flags=4099<UP,BROADCAST,MULTICAST>  mtu 1500
lo: flags=73<UP,LOOPBACK,RUNNING>  mtu 65536
```

### DIFF
--- a/bin/grep
+++ b/bin/grep
@@ -144,14 +144,14 @@ sub parse_args {
 
 	if (defined $opt{'f'}) {           # -f patfile
 		die("$Me: $opt{f}: is a directory\n") if ( -d $opt{f} );
-		open PATFILE, '<', $opt{f} or die qq($Me: Can't open '$opt{f}': $!);
+		open PATFILE, '<', $opt{f} or die qq($Me: Can't open '$opt{f}': $!\n);
 
 		# make sure each pattern in file is valid
 		while ( defined( $pattern = <PATFILE> ) ) {
 			chomp $pattern;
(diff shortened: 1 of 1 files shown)


bin/grep
```diff
@@ -144,14 +144,14 @@ sub parse_args {
 
 	if (defined $opt{'f'}) {           # -f patfile
 		die("$Me: $opt{f}: is a directory\n") if ( -d $opt{f} );
-		open PATFILE, '<', $opt{f} or die qq($Me: Can't open '$opt{f}': $!);
+		open PATFILE, '<', $opt{f} or die qq($Me: Can't open '$opt{f}': $!\n);
```
briandfoy, line 147: Not a big deal and I know you've been doing this elsewhere, but we can update these to lexical file handles when we run across them.

This doesn't block this patch, but I've just been in a vibe of not letting problems go past me at work, which is why my current task at work looks like a big PR and is taking too long. :)

briandfoy (approved):

briandfoy:
changes: silence a warning from grep reading stdin.

PR #745, opened 2024-09-26
grep: add -L option

mknos:
* GNU and BSD versions support -L which is the opposite of -l
* Using -L is easier than filtering grep -c output for 0
* Document that -L overrides -l and update usage string
* Bump version
```
%perl grep -c include ar a.c
ar:0
a.c:2
%perl grep -l include ar a.c
a.c
%perl grep -L include ar a.c
ar
```

### DIFF
--- a/bin/grep
+++ b/bin/grep
@@ -53,7 +53,7 @@ use File::Basename qw(basename);
 use File::Spec;
 use Getopt::Std;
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):
Brilliant. I probably knew that grep this feature, but it's great to be reminded.

PR #748, opened 2024-09-29
grep: retire -t extension

mknos:
* When running in recursive mode, standard grep is not required to support sorting of directory entries by modification time [1]
* The -t option does not exist in BSD, GNU, Solaris or Plan9 versions of grep
* Remove -t to follow BSD version more closely (bump version since now a usage string is printed for -t)

1. https://pubs.opengroup.org/onlinepubs/009604299/utilities/grep.html

### DIFF
--- a/bin/grep
+++ b/bin/grep
@@ -53,7 +53,7 @@ use File::Basename qw(basename);
 use File::Spec;
 use Getopt::Std;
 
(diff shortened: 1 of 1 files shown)


briandfoy:
changes: remove non-standard -t switch, since ordering dirs all the way down the line is not that easy

PR #749, opened 2024-09-30
grep: avoid shell code evaluation

mknos:
* When running grep -g or -u, the default code uses Term::Cap but this fails over to executing external tput command
* tput would be available on Linux/BSD but is not a consideration for Windows
* Feed the TERM environment variable to tput command in single quotes to prevent shell commands from being initiated though TERM
* To verify this I had to temporarily bypass the Term::Cap code on my Linux system
* No other instances of backticks, system() or exec() calls within grep
```
%TERM='linux;halt -p;' perl  grep -g include a.c   # with patch applied tput fails but system doesn't halt
tput: unknown terminal "linux;halt -p;"
tput: unknown terminal "linux;halt -p;"
#include <stdio.h>
#include <stdlib.h>
```

### DIFF
--- a/bin/grep
+++ b/bin/grep
@@ -199,8 +199,8 @@ sub parse_args {
 		else {                # if use of Term::Cap fails,
 			my $term = $ENV{'TERM'} || 'vt100';
 			( $SO, $SE ) = $opt{'g'}    # use tput command to get escapes
-				? ( `tput -T $term smso`, `tput -T $term rmso` )
-				: ( `tput -T $term smul`, `tput -T $term rmul` );
+				? ( `tput -T '$term' smso`, `tput -T '$term' rmso` )
+				: ( `tput -T '$term' smul`, `tput -T '$term' rmul` );
 			}
 		}
 
--- a/bin/grep
+++ b/bin/grep
@@ -51,6 +51,7 @@ use strict;
 
 use File::Basename qw(basename);
 use File::Spec;
+use File::Temp qw();
 use Getopt::Std;
 
 our $VERSION = '1.006';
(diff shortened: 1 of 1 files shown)


bin/grep
```diff
@@ -199,8 +199,8 @@ sub parse_args {
 		else {                # if use of Term::Cap fails,
 			my $term = $ENV{'TERM'} || 'vt100';
 			( $SO, $SE ) = $opt{'g'}    # use tput command to get escapes
-				? ( `tput -T $term smso`, `tput -T $term rmso` )
-				: ( `tput -T $term smul`, `tput -T $term rmul` );
+				? ( `tput -T '$term' smso`, `tput -T '$term' rmso` )
```
briandfoy, line 202: wouldn't we still be able to break this with single ticks in TERM, like `TERM="vt100' ; rm -rf / '"` (or something close)?

Maybe there are still some of the exploits I show in https://www.perl.com/article/quoting-the-shell/

mknos: Good point; I will add a new commit

bin/grep
```diff
@@ -121,6 +122,20 @@ EOF
 	}
 
 ###################################
+sub run_tput {
+	my $arg = shift;
+	my $term = $ENV{'TERM'} || 'vt100';
+
+	open my $oldout, '>&', STDOUT or die "$Me: can't save stdout: $!\n";
+	my $tmp = File::Temp->new;
+	open STDOUT, '>&', $tmp or die "$Me: can't redirect stdout: $!\n";
```
briandfoy, line 131: I probably would have used a pipe open here, but this is fine.

briandfoy (approved):

briandfoy:
changes: be more careful with handling the TERM env var

PR #751, opened 2024-10-03
expr: remove '==' alias for '='

mknos:
* GNU version supports == but it is not standardised [1] and other versions do not support it
* With patch applied, '==' results in syntax error, which is consistent with the OpenBSD version

1. https://pubs.opengroup.org/onlinepubs/9699919799/utilities/expr.html


briandfoy (approved):

briandfoy:
changes: == is no longer accepted as a synonym for = since this is not a standard option

PR #752, opened 2024-10-03
par: reject bad options

mknos:
* Change par to raise a fatal error if invalid options are passed, e.g. "perl par -X"
* Change the order or ARGV check to after getopts() call, so that an option is not counted incorrectly as a file argument, e.g. "perl par -S"


briandfoy (approved):

briandfoy:
changes: change to using getopts, so now unrecognized options complain

PR #754, opened 2024-10-04
what: exit(0) only for match

mknos:
* Similar to grep command, exit code 0 means at least one match occurred when looking at input files
* Unlike grep, exit code 1 is shared between no-match and error conditions (e.g. file could not be opened) [1]
* Continue to bail on the 1st open() failure, but now go through exit() instead of die()
* Difference discovered when testing against OpenBSD version

1. https://pubs.opengroup.org/onlinepubs/009696799/utilities/what.html

### DIFF
--- a/bin/what
+++ b/bin/what
@@ -15,16 +15,22 @@ use strict;
 
 use Getopt::Std qw(getopts);
 
+my $matches = 0;
 my %opt;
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: exit with 0 only if there are matches

PR #756, opened 2024-10-05
grep: remove code-generator

mknos:
* Define each of the different matching modes as its own closure within parse_args()
* All matching modes support multiple patterns
* The code is more readable because code doesn't get passed to eval() as a string
* The code in the closures is now able to fail at compile time, instead of deferring failure to runtime eval() call

### DIFF
--- a/bin/grep
+++ b/bin/grep
@@ -138,7 +138,78 @@ sub run_tput {
 }
 
 sub parse_args {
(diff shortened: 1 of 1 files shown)


briandfoy (approved):
oh gees, what a mess. I'm sure someone had some idea about how this could be interesting, but then it never really quite needed it.

PR #757, opened 2024-10-07
grep: allow repeated -e and -f

mknos:
* If -e and -f are not provided, the first argument to grep is a pattern
* If -e or -f is provided, the first argument to grep is a file
* Allow multiple -e and -f options to translate to multiple elements in the ```@patterns``` list
* This patch follows how GNU and BSD versions behave
* Bump version
* test1: "perl grep -e include -e return a.c" --> 2 patterns, one file argument
* test2: "echo include > P1 && echo return > P2 && perl grep -f P1 -f P2 a.c" --> 2 pattern files with one pattern each; one file argument
* test3: "grep return a.c" --> one (default) pattern argument,one file argument

### DIFF
--- a/bin/grep
+++ b/bin/grep
@@ -54,7 +54,7 @@ use File::Spec;
 use File::Temp qw();
 use Getopt::Std;
 
-our $VERSION = '1.006';
+our $VERSION = '1.007';
 
 $| = 1;                   # autoflush output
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: allow multiple -e or -f on the command line

PR #758, opened 2024-10-10
tsort: raise error for odd input tokens

mknos:
* The input "a" by itself results in a fatal error for BSD and GNU versions
* Previously this version ignored odd arguments, but that is contrary to the standard [1]
* Also, the standard does not suggest that each line contains only one pair of tokens
* When reading source code of OpenBSD version I discovered the input "a b b c b d d f c e" on one line is treated the same as for pairs on separate lines (GNU tsort behaves the same here)
* Accept multiple pairs per line for better compatibility, and remove the mention of the old limitation in the POD

1. https://pubs.opengroup.org/onlinepubs/9699919799/utilities/tsort.html

### DIFF
--- a/bin/tsort
+++ b/bin/tsort
@@ -11,7 +11,6 @@ License: perl
 
 =cut
 
-
 use strict;
 
 use File::Basename qw(basename);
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: raise error for an odd number of input tokens

PR #760, opened 2024-10-15
ed: w doesn't print byte count

mknos:
* Today I noticed that when starting ed without a file argument, writing a file with "w file.name" didn't print a byte count
* I confirmed that BSD and GNU versions of ed do print a byte count in this case
* When reading the code, the flag $SupressCounts should only be set when running ed in "script" mode (ed -s)
* The code setting the flag for no file argument is not correct (w command will fail unless a filename has either been provided directly as argument, or set earlier with f command)
* Bump version
```
%perl ed
a
this is ok
.
w some.file 
11
q
```

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -108,7 +108,7 @@ my $NO_QUESTIONS_MODE = 0;
 my $PRINT_NUM = 1;
 my $PRINT_BIN = 2;
 
-our $VERSION = '0.4';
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: w commands prints byte count when ed is started without a filename

PR #764, opened 2024-10-18
ed: basic repeated search

mknos:
* If I search for a pattern with "/hi" or "?bye" I want to repeat the search without re-entering the pattern (it might be much longer than "hi")
* Allow typing a single "/" or "?" to repeat search down or up respectively (this will fail if a previous search was not performed)
* GNU ed also allows repeating a search by typing empty pattern as "//" or "??", but that is not supported here for now


briandfoy (approved):

briandfoy:
changes: an empty search reuses the previous search

PR #765, opened 2024-10-19
ed: initial support for global (g) command

mknos:
* Basic usage of g command is to show lines matching a pattern
* More advanced use of g command would be to modify the buffer; this needs more work
* test1: g/text not found/p ---> no error is printed (consistent with BSD and GNU version)
* test2: g/include/p ---> print matching lines
* test3: g/include/n ---> prefix matching lines with line number
* test4: g/t/l ---> matching lines are listed in with escapes

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -13,8 +13,6 @@ License: gpl
 
 # What: A perl version of Unix ed editor.
 #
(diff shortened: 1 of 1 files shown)


briandfoy:
changes: add g command to grep a buffer

PR #766, opened 2024-10-21
cmp: helpful EOF messages

mknos:
* It is easy to provide extra descriptive output when files differ at EOF, as done by GNU cmp
* Shell scripts running cmp would normally use the -s flag and check the exit code, i.e. the text description shouldn't be depended on for portability
* Also convert $chunk_size global to a constant because it doesn't vary
```
%dd if=/dev/urandom bs=512 count=1 of=rnd
%cp rnd rnd2
%echo >> rnd2
%touch empty
%perl cmp rnd rnd2 # test1: non-empty files, eof on arg1
cmp: EOF on rnd after byte 512, in line 6
%perl cmp rnd2 rnd # test2: non-empty files, eof on arg2
cmp: EOF on rnd after byte 512, in line 6
%perl cmp empty rnd # test3: arg1 empty
cmp: EOF on empty which is empty
%perl cmp rnd empty # test4: arg2 empty
cmp: EOF on empty which is empty
```

### DIFF
--- a/bin/cmp
+++ b/bin/cmp
@@ -48,9 +48,9 @@ use constant EX_USAGE     => 2;
 use constant ST_INO  => 1;
 use constant ST_SIZE => 7;
 
-my $Program = basename($0);
+use constant READSZ => 10_000;
 
-my $chunk_size = 10_000;     # how many bytes in a gulp
+my $Program = basename($0);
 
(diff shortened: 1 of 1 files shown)


briandfoy:
changes: better messages when files differ at end of file

PR #767, opened 2024-10-22
ed: -s silences warning in edEdit()

mknos:
* Go with the BSD treatment of -s flag: allow e command to open a new file (replacing the line buffer) without warning if the buffer was modified
* This is more correct because in script mode (-s) ed is not being used interactively
* This difference was discovered when testing OpenBSD version


briandfoy:
changes: -s now also suppresses diagnostics when editing a file

PR #768, opened 2024-10-23
ed: allow repeated search with // and ??

mknos:
* In commit 78d54294099cfab729e156e8a0f123939da14179, repeat search was only implemented for "/" and "?"
* Add support for empty patterns specified as "??" and "//" (these may be followed by whitespace, but single "?" and "/" may not)
* test1: "/\n" --> repeat search down
* test2: "/ \n" --> start new search (down) for " "
* test3: "// \n" --> repeat search down
* test4: "?\n" --> repeat search up
* test5: "? \n"--> start new search (up) for " "
* test6: "?? \n --> repeat search up
* Entering //n doesn't seem to work for repeating search then issuing n command (this could be supported later)

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -1284,6 +1284,22 @@ Quit program
 
 Read named FILE into buffer
 
+=item /PATTERN
+
+Search for PATTERN in the buffer, starting from the current line.
+If no pattern is given the previous search is repeated.
+
+=item ?PATTERN
+
+Search backwards for PATTERN in the buffer, starting from the current line.
+If no pattern is given the previous search is repeated.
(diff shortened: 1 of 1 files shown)


briandfoy (changes requested):
Maybe it's time to add these to the docs under editor commands.

mknos:
Thanks for the reminder. I added a commit to document the search features, including global (g) command which doesn't support modifying the buffer at this time.

briandfoy (approved):

briandfoy:
changes: allow repeated search with // and ??

PR #769, opened 2024-10-24
file: -f option doesn't work properly

mknos:
* Non-standard option -f takes a path argument to read a list of paths to check
* Problem1: "file -f 0" is valid but incorrect truth check did not recognise -f
* Problem2: lines read from file are terminated by newlines, and these newlines were incorrectly being added into ARGV as part of the string (resulting in open() failure)
```
%cat 0
comm
ar
%perl file -f 0
comm: executable /usr/bin/perl script text
ar: executable /usr/bin/perl script text
```

### DIFF
--- a/bin/file
+++ b/bin/file
@@ -123,12 +123,16 @@ GetOptions(
     ) or usage();
 
 # the names of the files are in $fileList.
-if ($fileList) {
+if (defined $fileList) {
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: -f works properly by checking the definedness of its argument instead of truthiness, and chomps the lines it reads.

PR #770, opened 2024-10-25
ed: basic support for inverted global search (v)

mknos:
* Implement inverted global command 'v' with the same limitations as 'g' command
* Document the 'v' command in pod, and remove old comment suggesting to add it
* Avoid incorrectly printing $Prompt when running 'g' command (found when running P before g)
* Global search should not proceed from the current line (found when testing against GNU ed)
* Bump version
* test1: "v/include/n" --> list lines not matching "include", starting from line 1
* test2: jump to line 5 then repeat test1 --> search items are displayed in the same order as before, irrespective of current line position

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -49,7 +49,6 @@ License: gpl
 #        - Implement the following commands from the v7 docs:
 #                k - mark
 #                u - undo
-#                v - global command "inVerted"
 #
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: add v command for inverted searches

PR #772, opened 2024-10-28
ed: retire old command parser

mknos:
* Replace large regex in edParse() with two calls to a getAddr() function
* edParse() is now more friendly to step through with a debugger
* CalculateLine() is no longer called, so delete it
* getAddr() is currently missing the regex-pattern address mode; it will be added soon
* The large regex did not always behave correctly when a search pattern was given as an address
* Bump version


briandfoy (approved):
So much cleaner

PR #773, opened 2024-10-29
ed: re-add support for /re/ address-mode

mknos:
* Standard ed allows a regex to be placed in front of a command
* Previously CalculateLine() supported this but the function was replaced by getAddr()
* Add a new condition in getAddr() to allow resolving /re/ command prefix to a line number
* Be careful to support repeated search for empty pattern
* test1: /include/n  ---> search for /include/, resolve line number and run n command
* test2: //l ---> repeat search using saved /include/ pattern, then run l command on matching line
* test3: /return/  ---> no command given with pattern still works
* test4: ```/a\/b/p``` --> search pattern includes escaped slash; run p command on matching line

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -958,6 +958,25 @@ sub getAddr {
         $n = $CurrentLineNum;
     } elsif (s/\A\$//) { # '$' == max line
         $n = maxline();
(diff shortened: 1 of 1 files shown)


briandfoy:
changes:  re-add support for /re/ address-mode

PR #774, opened 2024-10-30
ed: support ?re? address mode

mknos:
* ?re? is a valid prefix which searches backward
* Tested against GNU and OpenBSD versions
* test1: ?perl?n ---> search for previous instance of "perl" then run command n
* test2: ??l ---> search for previous instance of saved pattern "perl" then run command l
* test3: ```?\??p``` ---> search backwards for literal "?" then run command p
* test4: ```?\/\/?``` ---> search backward for literal "//" then show result


briandfoy (approved):

briandfoy:
changes: support ?re? address mode

PR #775, opened 2024-10-30
ed: fix no-match showing multiple error prompts

mknos:
* Delete old code for handling repeated search since getAddr() now supports it
* edParse() must return true for a valid command, so defer to input() (its caller) to distinguish NoMatch and NoSavedPattern errors
* getAddr() may be called twice, and NoMatch/NoSavedPattern error could occur either time
* For "g" command, the NoMatch condition is not an error (no action is performed)
* Fix an issue with parsing 2 addresses: "2,n" is equivalent to "2n", not "2,$n"
* test1: "//l" ---> error: no previous pattern
* test2: ",/main/n" ---> run command n for lines ranging from 1 to the 1st match of /main/
* test3: "?NO LINE MATCHES THIS!!!?" ---> error: no match
* test4: "2,n" ---> just line 2
* test5: ",2n" ---> lines 1 to 2
* test6: ",n" ---> lines 1 to $Max

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -235,7 +235,13 @@ sub input {
     # sanity check addresses
     foreach my $ad (@adrs) {
         next unless defined $ad;
-        if ($ad > maxline() || $ad < 0) {
+        if ($ad == -1) {
+            edWarn(E_NOMATCH);
+            return;
+        } elsif ($ad == -2) {
+            edWarn(E_NOPAT);
+            return;
+        } elsif ($ad > maxline() || $ad < 0) {
             edWarn(E_ADDRBAD);
             return;
         }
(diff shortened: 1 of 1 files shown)


bin/ed
```diff
@@ -235,7 +235,13 @@ sub input {
     # sanity check addresses
     foreach my $ad (@adrs) {
         next unless defined $ad;
-        if ($ad > maxline() || $ad < 0) {
+        if ($ad == -1) {
```
briandfoy, line 238: Maybe some constant names for these magic numbers?

briandfoy (changes requested):

bin/ed
mknos: Good idea, added.

briandfoy (approved):

PR #777, opened 2024-11-01
ed: allow address ranges with g command

mknos:
* When testing "g" command against BSD version I discovered 1- and 2-address prefix is accepted
* Implement this by looking at ```@adrs``` in edSearchGlobal(); address list has been populated earlier in edParse()
* test1: g/include/p ---> process lines 1 to $Max
* test2: 1g/include/p ---> process only line 1
* test3: 1,2g/include/p ---> process lines 1 to 2
* test4: ,2g/include/p ---> same as test3

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -1033,8 +1033,24 @@ sub edSearchBackward {
 sub edSearchGlobal {
     my($pattern, $invert) = @_;
 
+    my $end = maxline();
+    my $start = 1;
+    if (defined $adrs[0]) {
+        if ($adrs[0] == 0) {
+            edWarn(E_ADDRBAD);
+            return;
+        }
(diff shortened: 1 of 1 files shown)


briandfoy:
I don't have time to try this right now, but does the lone comma work too (implicit first to last)?

    ,g/pattern/p

mknos:
Yes, the lone comma does the same as running "g" command without any address, processing lines 1 to $Max. This is consistent with GNU and BSD versions.
In this version, ```@adrs``` is populated by edParse() for the lone-comma case before calling edSearchGlobal().
The default values at top of edSearchGlobal() handle ```@adrs``` being empty, which happens when running "g" with no prefix.

PR #782, opened 2024-11-05
ed: support % for all-addresses

mknos:
* GNU ed treats '%' as an alias for ',' and allows it to be used either alone or as a delimiter between addresses
* OpenBSD ed only allows '%' to be used alone,  with the meaning of "all addresses"
* Go with the BSD meaning of '%' and implement lone-percent prefix in command parser
* test1: %n ---> same as ,n
* test2: %s/float/double/ ---> perform one substitution per line for all lines in editor buffer
* test3: 1%2p ---> invalid
* test4: 1%p ---> invalid
* test5: %2p ---> invalid

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -62,6 +62,7 @@ use Getopt::Std qw(getopts);
 
 use constant A_NOMATCH => -1;
 use constant A_NOPAT   => -2;
+use constant A_ALL     => -3;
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: % takes OpenBSD meaning of "all addresses"; it can only be used alone

PR #783, opened 2024-11-05
cut: fix usage of -b address range

mknos:
* Range argument for -b is expressed as "a-b"
* Either a or b can be omitted, but not both
* "a-" means "from a to end"
* "-b" means "from 1 to b"
* Previously "a-" usage was not handled properly
* Found when testing against BSD version
* test1: "echo abcdefg | perl cut -b 2" ---> b
* test2: "echo abcdefg | perl cut -b 2-3" ---> bc
* test3: "echo abcdefg | perl cut -b -3" ---> abc
* test4: "echo abcdefg | perl cut -b 3-" ---> cdefg

### DIFF
--- a/bin/cut
+++ b/bin/cut
@@ -59,25 +59,46 @@ sub handle_b {
     my $spec = shift;
     my @list = split /,/, $spec;
 
+    my $to_end;
     my @cols;
     foreach my $item (@list) {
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: correctly handle "start, no end" (`n-`) range in `-b`

PR #784, opened 2024-11-05
du: runtime error for bad environment value

mknos:
* Add validation that $blocksize must be a positive integer
* Setting BLOCKSIZE="-0" would result in a division by zero
* Possibly du doesn't need to access any environment variables, but for now keep BLOCKSIZE

### DIFF
--- a/bin/du
+++ b/bin/du
@@ -36,6 +36,9 @@ $blocksize = $ENV{BLOCKSIZE} if $ENV{BLOCKSIZE}; # use environment if present
 getopts('HLPacklrsx') or usage();
 
 $blocksize = 1024 if $opt_k;
+if ($blocksize =~ m/[^0-9]/ || $blocksize == 0) {
+  die "$0: unexpected block size: $blocksize\n";
+}
 
 if ($opt_a && $opt_s) {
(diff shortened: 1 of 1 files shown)


bin/du
```diff
@@ -36,6 +36,9 @@ $blocksize = $ENV{BLOCKSIZE} if $ENV{BLOCKSIZE}; # use environment if present
 getopts('HLPacklrsx') or usage();
 
 $blocksize = 1024 if $opt_k;
+if ($blocksize =~ m/[^0-9]/ || $blocksize == 0) {
+  die "$0: unexpected block size: $blocksize\n";
```
briandfoy, line 40: du on archlinux returns `1`, FreeBSD returns 64 (for `-B`), and so on. I'd prefer some value that isn't `die`'s 255.

briandfoy (changes requested):
Maybe a better exit value

bin/du
mknos: Ok, I've added a new commit and there are no longer any calls to die() in the du command.

briandfoy (approved):

briandfoy:
changes: properly handle illegal BLOCKSIZE env values

PR #786, opened 2024-11-06
ed: implement global substitute

mknos:
* The input g/search1/s/search2/replace/ allows search1 to filter the lines to process, then search2 determines the text to replace per line
* Don't return in the condition for g & v commands; this allows the subsequent condition to determine the command to be run globally
* Using rindex() to find a delimiting '/' was incorrect because s/// confuses things with its own '/'
* Adapt some code from getAddr() to find delimiter of search1 regex
* test1: ```g/\//s/$/ NL/```   ---> append the text " NL" to all lines matching literal '/'
* test2: ```v/\//d```   ---> delete all lines not matching literal '/'

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -109,7 +109,7 @@ my $NO_QUESTIONS_MODE = 0;
 my $PRINT_NUM = 1;
 my $PRINT_BIN = 2;
 
-our $VERSION = '0.11';
+our $VERSION = '0.12';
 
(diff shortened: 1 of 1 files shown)


briandfoy:
changes: implement global substitute: g/search1/s/search2/replace/ allows search1 to filter the lines to process, then search2 determines the text to replace per line

PR #790, opened 2024-11-08
ed: make wq shortcut work again

mknos:
* Previously typing wq was supported, but recent changes in edParse() broke it
* Command suffix processing goes before check for a space, then optional filename argument
* This time document wq in pod


briandfoy (approved):

briandfoy:
changes: make wq shortcut work again after previous fix changed it

PR #792, opened 2024-11-09
ed: implement ';' addressing mode

mknos:
* ";" by itself is taken to mean current address to end, i.e. ".,$"
* "A1;A2" expression is the same as "A1,A2"
* ";A2" expression works like ".,A2"
* I tested the following combinations...

,p   ---> all lines
%p   ---> all lines
;p   ---> current line to end
1,p  ---> only line 1
1%p  ---> not supported
1;p  ---> only line 1
1,2p ---> range 1-2
1%2p ---> not supported
1;2p ---> range 1-2
,2p  ---> range 1-2
%2n  ---> range 1-2
;20p ---> range current line to 20 (error if current>20)

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -62,8 +62,7 @@ use Getopt::Std qw(getopts);
 
 use constant A_NOMATCH => -1;
 use constant A_NOPAT   => -2;
-use constant A_ALL     => -3;
-use constant A_PATTERN => -4;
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes:  implement ';' addressing mode

PR #793, opened 2024-11-11
ed: retire non-standard debug flag -d

mknos:
* This version of ed is different to the C versions because it's trivial to run it under the perl debugger, e.g. setting a breakpoint in edParse() or input()
* The debug info produced by running "ed -d" is limited, and other versions of ed don't provide an equivalent option
* Sync usage string + SYNOPSIS
* Add symbolic exit codes as done in other scripts

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -79,6 +79,9 @@ use constant E_PATTERN => 'invalid pattern delimiter';
 use constant E_NOMATCH => 'no match';
 use constant E_NOPAT   => 'no previous pattern';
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):
I don't particularly agree with this change in principle, but if the `-d` really doesn't help because it has low instrumentation, yeah, rip it out. If there was good instrumentation, I'd probably go the other way.

briandfoy:
changes: remove -d (debugging) switch since the debugging output isn't that helpful

PR #794, opened 2024-11-11
what: compatible output text for match

mknos:
* My Linux system doesn't have a "what" command
* The OpenBSD version prints each match on a separate line with a tab character prefix
* The standards document is clear about ```printf("\t%s\n", id_string)```, so follow that

1. https://pubs.opengroup.org/onlinepubs/9699919799/utilities/what.html


briandfoy (approved):

briandfoy:
changes: line prefix is a tab instead of eight spaces

PR #795, opened 2024-11-11
grep: search binary files

mknos:
* When testing against OpenBSD and GNU versions, the meaning of -a is to list the matching lines of a binary file 
* Without -a, binary files should be searched but only a "Match" message is printed
* Change the current meaning of -a to make this version more compatible
* Update pod manual and description in usage()
```
%perl grep BASH /bin/bash
grep: /bin/bash: binary file matches
%echo $?
0
%perl grep BASH /bin/echo
%echo $?
1
%perl grep -a BASH /bin/bash
/usr/share/bashdb/bashdb-main.inccannot start debugger; debugging mode disabledrbash/bin/shI have no name!??host??run_one_command-cENVBASH_ENVHISTFILEcannot set uid to %d: 
...
%echo $?
0
%perl grep -a BASH /bin/echo
%echo $?
1
```

### DIFF
--- a/bin/grep
+++ b/bin/grep
@@ -54,7 +54,7 @@ use File::Spec;
 use File::Temp qw();
 use Getopt::Std;
 
-our $VERSION = '1.008';
+our $VERSION = '1.009';
 
 $| = 1;                   # autoflush output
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: -a (binary files) changed to act like the OpenBSD/GNU by printing the "line" that matched

PR #797, opened 2024-11-12
ed: support '^' address

mknos:
* For greater compatibility with BSD, either '-' or '^' can be specified to mean CurrentLine - 1
* GNU ed 1.19 doesn't support '^'; I didn't test if v1.20 supports it
* Standards document mentions '^' is considered a historical feature; however POSIX does not prohibit it [1]

1. https://pubs.opengroup.org/onlinepubs/9699919799/utilities/ed.html


briandfoy (approved):

briandfoy:
changes: support '^' addressing

PR #798, opened 2024-11-12
chgrp: accept group ID argument

mknos:
* The "group" argument to chgrp can be either a name or an ID
* When testing against GNU and OpenBSD I found that this version did not accept an ID argument
* Allow numeric ID to be passed to chown()
* Also fix a problem where uid=0 from stat() was treated as stat() failure; it is normal for uid=0 to be root


briandfoy (approved):

briandfoy:
changes: accept group ID argument

PR #799, opened 2024-11-12
grep: don't automatically decompress

mknos:
* When run on a binary file, grep should search it as a binary file
* Regular grep should not decompress a file (this version was decompressing based on file extension)
* Adopt the -Z flag from OpenBSD which makes grep behave as zgrep [1]
* Sync POD manual

1. http://man.openbsd.org/grep#Z

```
%gzip -9k ar # create ar.gz
%perl grep ar ar.gz # original filename stored in binary
grep: ar.gz: binary file matches
%perl grep perl ar.gz # no match
%perl grep -Z perl ar.gz # match after decompression
#!/usr/bin/perl
License: perl
```

### DIFF
--- a/bin/grep
+++ b/bin/grep
@@ -54,7 +54,7 @@ use File::Spec;
 use File::Temp qw();
 use Getopt::Std;
 
-our $VERSION = '1.009';
+our $VERSION = '1.010';
 
 $| = 1;                   # autoflush output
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: add -Z to operate in zgrep mode (don't automatically decompress everything)

PR #800, opened 2024-11-12
mail: trim leading spaces from command input

mknos:
* For compatibility with OpenBSD mail program, "   q" should be interpreted as "q"
* While here, fix the name of the "visual" command (aka "v") in the pod manual (entering "visal" does nothing and it's invalid on OpenBSD too)


briandfoy (approved):

briandfoy:
changes: trim leading spaces from command input

PR #801, opened 2024-11-12
ls: avoid invalid closedir()

mknos:
* When temporarily enabling warnings.pm I observed a warning in DirEntries(): closedir() attempted on invalid dirhandle $dh
* opendir() was always called even if the file was not a directory
* Delay the opendir() call until we know that we don't want to list a regular file
* opendir() should not happen if running "ls -d", or if the argument to DirEntries() is not a directory

### DIFF
--- a/bin/ls
+++ b/bin/ls
@@ -150,15 +150,18 @@ sub DirEntries {
 	my @Entries = ();	# entries in original order
 	my $Name = "";		# entry name
 
-	if (!opendir($dh, $_[0]) || exists($Options{'d'})) {
+	if (exists $Options{'d'} || ! -d $_[0]) {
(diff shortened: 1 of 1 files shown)


bin/ls
```diff
@@ -150,9 +150,8 @@ sub DirEntries {
 	my @Entries = ();	# entries in original order
 	my $Name = "";		# entry name
 
-	if (!opendir($dh, $_[0]) || exists($Options{'d'})) {
```
briandfoy, line 153: yeah, this line makes no sense

briandfoy (approved):

briandfoy:
changes: don't automatically opendir everything

PR #803, opened 2024-11-12
rmdir: better error reporting

mknos:
* The old error message is not specific enough to be helpful
* Allow rmdir() to fail instead of performing file tests first
* $! will specifically say whether a file argument does not exist, or is not a directory


briandfoy (approved):

briandfoy:
changes: test that an argument is a directory before trying to remove it (better error message)

PR #804, opened 2024-11-12
chown: allow numeric UID argument

mknos:
* This is similar to the patch for chgrp command
* Fall back to treating numeric argument as a uid if username resolution failed
* The code surrounding stat() had the same bug as for chgrp, i.e. existing uid=0 was treated as a failed stat()


briandfoy (approved):

briandfoy:
changes: allow numeric UID argument

PR #806, opened 2024-11-13
apply: improve parsing of -a option

mknos:
* When testing against OpenBSD, the -a option argument can be given with or without a leading space
* Allow this version to accept the "-a c" form to avoid confusion
* Sync usage string with BSD version: "[-a c]" instead of [-ac] which could be confused as grouped options -a and -c
* If a bad option was given, e.g. -x, print it before the usage string
* If the argument to -a is >1 character, reject it and terminate the program
* test1: perl apply -a'!' 'echo HELLO, !1' e*.c # say hello to my e files
* test2: perl apply -a '!' 'echo HELLO, !1' e*.c # same
* test3: perl apply 'echo HELLO, %1' e*.c # same, without -a
* test4: perl apply -a 'oops' 'echo HELLO, !1' e*.c # invalid

### DIFF
--- a/bin/apply
+++ b/bin/apply
@@ -21,7 +21,7 @@ use constant EX_SUCCESS => 0;
 use constant EX_FAILURE => 1;
 
 my $Program = basename($0);
-my ($VERSION) = '1.4';
+my ($VERSION) = '1.5';
 
 my $argc  = 1;
 my $debug = 0;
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: allow a space after -a so it isn't confused with bundled options

PR #809, opened 2024-11-13
grep: replace -1 with -m NUM

mknos:
* The -1 flag meant to limit searches to one match per file
* The -1 flag is incompatible with GNU and BSD grep because they interpret -1 as -C 1, i.e. show one line of context around each match
* Implement the common way to limit searches by adding -m (maximum match) option
* Allow -m0 because this is supported in the other versions
* test1: perl grep -1 pat file1 file2 ---> no longer supported
* test2: perl grep -m 2 pat file1 file2 ---> two matches per file

### DIFF
--- a/bin/grep
+++ b/bin/grep
@@ -54,7 +54,7 @@ use File::Spec;
 use File::Temp qw();
 use Getopt::Std;
 
-our $VERSION = '1.010';
+our $VERSION = '1.011';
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):
Good catch

briandfoy:
changes: Remove -1 switch. Use -m 1 instead, for GNU compatibility

PR #810, opened 2024-11-13
grep: implement -A context option

mknos:
* Show non-matching context lines after a matching line
* A new match may occur within the context lines being displayed; this resets the context counter unless the match limit (-m value) was already reached
* Display a line of "--" to indicate the end of the context block
* If -A is used with -n, follow line number with '-' instead of ':' to indicate context lines
* When -A is used with -m, terminate the search at the end of the context and not the exact line triggering match limit

```
%perl grep -n -A 8 perl ar # test1: context bleeds into the line8 match
1:#!/usr/bin/perl
2-
3-=begin metadata
4-
5-Name: ar
6-Description: create and maintain library archives
7-Author: dkulp
8:License: perl
9-
10-=end metadata
11-
12-=cut
13-
14-use strict;
15-
16-use POSIX qw(strftime);
--
%perl grep -n -A 8 -m 1 perl ar # test2: line8 match doesn't count due to -m1 limit
1:#!/usr/bin/perl
2-
3-=begin metadata
4-
5-Name: ar
6-Description: create and maintain library archives 
7-Author: dkulp
8:License: perl
9-
--
%perl grep -A 4 perl ar # test3: no -n
#!/usr/bin/perl

=begin metadata

Name: ar
--
License: perl

=end metadata

=cut
--
```

### DIFF
--- a/bin/grep
+++ b/bin/grep
@@ -54,7 +54,7 @@ use File::Spec;
 use File::Temp qw();
 use Getopt::Std;
 
-our $VERSION = '1.011';
+our $VERSION = '1.012';
 
 $| = 1;                   # autoflush output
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: implement -A for context after a match

PR #812, opened 2024-11-15
ed: basic support for mark (k) command

mknos:
* Mark is a standard feature in ed and it's not very much code to support it
* Resolving a mark that doesn't exist results in Invalid Address error
* 1,2kx ---> mark x as 2
* 1kx ---> mark x as 1
* kx ---> mark x as CurrentLine (default)
* 'x,$n ---> run n command from marked line x to end of buffer

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -47,7 +47,6 @@ License: gpl
 #
 # Todo:
 #        - Implement the following commands from the v7 docs:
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: basic support for mark (K) command

PR #814, opened 2024-11-15
chmod: exit(1) if a file fails

mknos:
* A failed chmod() is not a fatal error; the program will proceed to the next file argument but will exit(1) to indicate an error occurred
* Do the same in modify_file() if a file doesn't exist (this helps exitcode match the GNU version)
* Also introduce a regular usage() function

### DIFF
--- a/bin/chmod
+++ b/bin/chmod
@@ -15,28 +15,13 @@ use strict;
 
 use File::Basename qw(basename);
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: exit with 1 if a file fails

PR #815, opened 2024-11-17
grep: add -I flag

mknos:
* BSD and GNU versions have an -I option for ignoring binary files, i.e. treating binary files as if they don't match
* test1: perl grep BASH /bin/bash ---> match, suppress text
* test2: perl grep -a BASH /bin/bash ---> match, show text
* test3: perl grep -q BASH /bin/bash ---> match, suppress all output
* test4: perl grep -I BASH /bin/bash text.txt ---> match on text.txt but not /bin/bash

### DIFF
--- a/bin/grep
+++ b/bin/grep
@@ -54,7 +54,7 @@ use File::Spec;
 use File::Temp qw();
 use Getopt::Std;
 
-our $VERSION = '1.013';
+our $VERSION = '1.014';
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: add -I (capital i) switch to ignore binary files

PR #816, opened 2024-11-17
grep: make space after -e or -f optional

mknos:
* BSD and GNU versions do not require a space between -e/-f and the option argument
* This version did not raise an error, but option arguments without a space resulted in incorrect behaviour
* Remove -e and -f from getopts() option string because they are handled earlier
* test1: perl grep -e0 -e 1 ar ---> patterns 0 and 1 in file ar
* test2: grep -fwordlist.txt.old -f pat1 xargs ---> patterns from file wordlist.txt.old and pat1, searching against file xargs

### DIFF
--- a/bin/grep
+++ b/bin/grep
@@ -54,7 +54,7 @@ use File::Spec;
 use File::Temp qw();
 use Getopt::Std;
 
-our $VERSION = '1.014';
+our $VERSION = '1.015';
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: make space after -e or -f optional

PR #817, opened 2024-11-17
grep: remove paragraph mode

mknos:
* Remove options -p and -P, which don't correspond with any other version of grep
* Standard grep would require formatting paragraph onto one line first before feeding it to grep


briandfoy (approved):

briandfoy:
changes: remove -p and -P for paragraph mode, which was a non-standard perl extension

PR #818, opened 2024-11-18
du: exit code 1 on error

mknos:
* Failure to access one file argument is considered a non-fatal error which deserves an exit(1) at end of program
* Tested against GNU version
```
%perl du -k  a aaaaaaaa
66	a
du: cannot access 'aaaaaaaa': No such file or directory
%echo $?
1
```

### DIFF
--- a/bin/du
+++ b/bin/du
@@ -53,11 +53,13 @@ unless (@ARGV) {
   push @ARGV, '.';
 }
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes:  exit code 1 on error

PR #819, opened 2024-11-18
uniq: one input file

mknos:
* Standards document explains uniq has two optional file arguments, input and output [1]
* GNU and BSD versions follow this; do the same and stop treating subsequent file arguments as extra input files
* test1: printf "a\na\nb\nc\nc\n" | perl uniq ---> implicit stdin and stdout
* test2: perl uniq a.c ---> read a.c, write to stdout
* test3: perl uniq a.c b.c ---> read a.b, write to b.c
* test4: perl uniq A B C ---> invalid, too many args

1. https://pubs.opengroup.org/onlinepubs/009696799/utilities/uniq.html

### DIFF
--- a/bin/uniq
+++ b/bin/uniq
@@ -84,7 +84,7 @@ if (defined $infile) {
 }
 if (defined $outfile) {
   unless (open $out, '>', $outfile) {
-    warn "$0: failed to open '$infile': $!\n";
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: arguments are now exactly one input file and optionally one output file

PR #820, opened 2024-11-18
grep: add -R alias for -r

mknos:
* On *BSD, there's no difference between grep -r and grep -R, so adding -R in this version allows for greater compatibility
* Also add missing validation: $pattern can be undefined when shifting off ARGV, and it's a fatal error if we get no patterns (found with "perl grep -r")


briandfoy (approved):

briandfoy:
changes: make -R a synonym for -r

PR #821, opened 2024-11-18
grep: no match for empty pattern file

mknos:
* Testing against old commit 504caa95c6979a507504365047af3dd13d30fac8, "grep -f empty" resulted in exit(1), i.e. no match
* A recent commit introduced a bug where ARGV would be shifted incorrectly if an empty file was passed as argument to -f
* Follow GNU grep and bypass the search in this case because we know there cannot be any matches
```
%touch empty && perl grep -f empty ar
%echo $?
1
```

### DIFF
--- a/bin/grep
+++ b/bin/grep
@@ -54,7 +54,11 @@ use File::Spec;
 use File::Temp qw();
 use Getopt::Std;
 
-our $VERSION = '1.017';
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: return no match exit code for an empty pattern file used with -f

PR #822, opened 2024-11-18
chown: allow numeric group argument

mknos:
* Previously the usage "chown USER file" was changed to allow numeric user ID for USER
* When testing against GNU version, the usage "chown USER:GROUP file" should handle a group ID or a group name
* test1: perl chown 0:0 file ---> user=root, group=root on my Linux system
* test2: perl chown root:8 ---> also valid, where gid 8 resolves to "mail" here


briandfoy (approved):

briandfoy:
changes: allow for numeric group in user:group notation

PR #823, opened 2024-11-19
mail: more reliably find vi command

mknos:
* On Android I have Termux, and vi is installed as /data/data/com.termux/files/usr/bin/vi
* Remove the assumption that vi is installed under /usr/bin
* With this patch, perl will look for vi within $PATH
* Tested on OpenBSD and Termux/Android
* Environment variable $VISUAL still takes precedence if it is set


briandfoy (approved):

briandfoy:
changes: use VISUAL for the path to vi, or look for vi in the PATH

PR #827, opened 2024-11-20
cp -i: continue on prompt

mknos:
* cp can copy multiple files to a destination
* It is not an error for the user to type "n" to not clobber an existing file when running "cp -i"
* Previously if I type "n", cp exits with a failure code and ignores the remaining files I want to copy
* This patch is for the scenario where an external cp command is not found

```
%mkdir dest && perl cp -v a.c ar dest && perl cp -i a.c ar dest
a.c -> dest/a.c
ar -> dest/ar
overwrite dest/a.c? (y/n [n]) [n] n
overwrite dest/ar? (y/n [n]) [n] y
%echo $?
0
```


briandfoy (approved):

briandfoy:
changes: cp -i will continue to next file if there is an error.

PR #828, opened 2024-11-21
chmod: check for readlink() failure

mknos:
* In "chmod -R -L" case, check readlink() return value before use
* If readlink() failed (which shouldn't happen), set exit code for failure
* Reduce readlink() calls from 2 to 1 per file argument to "chmod -R -L"

```
%ls -l what whatlnk
-rwxr-xr-x 1 pi pi 3123 Nov 12 07:18 what
lrwxrwxrwx 1 pi pi    4 Nov 21 12:28 whatlnk -> what
%perl chmod -R -L 400 whatlnk
%ls -l what whatlnk
-r-------- 1 pi pi 3123 Nov 12 07:18 what
lrwxrwxrwx 1 pi pi    4 Nov 21 12:28 whatlnk -> what
```

### DIFF
--- a/bin/chmod
+++ b/bin/chmod
@@ -19,7 +19,7 @@ use constant EX_SUCCESS => 0;
 use constant EX_FAILURE => 1;
 
 my $Program = basename($0);
-my $VERSION = '1.4';
+my $VERSION = '1.5';
 
 my $rc = EX_SUCCESS;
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: check for redline failures and exit with failure code when they fail

PR #830, opened 2024-11-22
ed: m command accepts address argument

mknos:
* Argument parsing within edMove() was a bit broken
* edMove() implements the "t" and "m" commands
* I noticed that bogus values for $args[0] were being accepted, e.g. n > maxline()
* The argument is supposed to be any valid address, not just an address specified as a number
* getAddr() already contains the functionality we want, so call it

Tests:
1,2m/main/ ---> append selected lines after regex match
1,2m/mainxxxxxx/ ---> nomatch error
1,2m// ---> use saved pattern for search (fails if no pattern was saved yet)
1,2m. ---> append lines after current address (explicit)
1,2m ---> append lines after current address (default)
1,2m$ ---> append lines after last line
1,2m'x ---> append lines after marked line x (fails if nothing was marked for x yet)
24,25m0 --> move selected lines to start of editor buffer

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -113,7 +113,7 @@ my $NO_QUESTIONS_MODE = 0;
 my $PRINT_NUM = 1;
 my $PRINT_BIN = 2;
 
-our $VERSION = '0.19';
+our $VERSION = '0.20';
 
 my @ESC = (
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: m command accepts address argument

PR #833, opened 2024-11-23
mail: add --version

mknos:
* Declare VERSION_MESSAGE() as done in other scripts which use getopts()
* It is not necessary to print a splash message when the program starts; displaying the mailbox is enough
* Printing os in version message is not required (English.pm was only used for $OSNAME which disappears in this patch)
* The -v option is for toggling debug information, and it doesn't change

### DIFF
--- a/bin/mail
+++ b/bin/mail
@@ -633,7 +633,6 @@ EDLOOP: {
 1;
 package main;
 
-use English;
 use File::Temp;
 use Getopt::Std;
 use vars qw($opt_f $opt_s $opt_c $opt_b $opt_v);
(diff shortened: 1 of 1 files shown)


bin/mail
```diff
@@ -1037,6 +1035,11 @@ sub Batch {
 	$mailer->send($message);
 }
 
+sub VERSION_MESSAGE {
+	print "$0 version $VERSION\n";
```
briandfoy, line 1039: Let's fiddle with `$0` to get the basename for the program like we've done in other programs.

briandfoy (changes requested):

bin/mail
mknos: Ok, I've added $Program here variable for consistency

PR #835, opened 2024-11-25
fold: remove assert in adj_col()

mknos:
* When running fold without -b flag, adj_col() is used---called through fold_file()
* "fold -b" is a different algorithm and does not use adj_col()
* Remove the $Byte_Only check in adj_col() because this can never happen
* Also make the usage() function more standard by not taking a parameter


briandfoy (approved):

briandfoy:
changes: improve usage message (no longer takes a parameter)

PR #836, opened 2024-11-25
expr: allow unary plus

mknos:
* When testing OpenBSD expr, +n and -n numbers are allowed, so permit them here
* I didn't test NetBSD version, but since it calls strtoll() I suspect it also permits +n and -n [1]
* Test: ```perl expr 2 \* \( -2 + +43 \)``` ---> 82

http://cvsweb.netbsd.org/bsdweb.cgi/src/bin/expr/expr.y?annotate=1.46

### DIFF
--- a/bin/expr
+++ b/bin/expr
@@ -36,7 +36,7 @@ if (scalar(@ARGV) == 0) {
 
 # check that it is a number
 sub num($) {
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: numbers can start with +

PR #837, opened 2024-11-25
install: -m0 option is valid

mknos:
* I might want to install something but make it temporarily not accessible --- "install -m 0 file dir"
* Numeric file mode 0 should be passed to chmod() but it was being ignored
* Problem1: modify_file() assumes $mode is not zero
* Problem2: install_files() assumes $opt{m} is not zero
* Apply same logic in install_dirs() because that seems wrong too
* In install_files(), calculating the symbolic file mode is not required in the "non-unix" case, where modify_file() would never be called (I noticed this when comparing install_dirs())

### DIFF
--- a/bin/install
+++ b/bin/install
@@ -17,7 +17,7 @@ use strict;
 use File::Basename qw(basename dirname);
 use Getopt::Std qw(getopts);
 
-our $VERSION = '1.4';
+our $VERSION = '1.5';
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: allow mode to be 0 (make thing temporarily unavailable even when installed)

PR #841, opened 2024-11-27
install: fix absolute symbolic permissions

mknos:
* When using symbolic permissions with '=', the resulting file permissions weren't right; here's what I found...
* mod() should be passed the original file to stat(), but it was being passed the new target file after File::Copy::copy() created it
* Base permissions for "absolute" symbolic permissions (i.e. those set with '=') are zero, so it is not correct to stat() the original file in this case
* The statement clearing bits in ```%perms``` for operator '=' in loop should only clear bits for $who the current iteration looks at

```
%perl install -m u=rwx,g=rx,o=x comm a/comm # -m 751
%perl install -m u=rwx,g=r comm a/comm # -m 740
%perl install -m 'a=' comm a/comm # -m 0
```

### DIFF
--- a/bin/install
+++ b/bin/install
@@ -17,7 +17,7 @@ use strict;
 use File::Basename qw(basename dirname);
 use Getopt::Std qw(getopts);
 
-our $VERSION = '1.6';
+our $VERSION = '1.7';
 my $Program = basename($0);
 
 sub VERSION_MESSAGE {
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: handle symbolic permissions correctly

PR #842, opened 2024-11-27
hangman: consistently handle eof

mknos:
* If you reach the end of a game and see the PLAY AGAIN prompt, eof will cause the program to terminate
* Do the same if eof occurs after the GUESS A LETTER prompt; the NetBSD version handles this in getguess() [1]
* test: echo sword > wordlist.txt  && echo hi | perl hangman # now terminates 

1. http://cvsweb.netbsd.org/bsdweb.cgi/src/games/hangman/getguess.c?annotate=1.10


briandfoy (approved):

briandfoy:
changes: consistently handle eof when there is not a guess

PR #844, opened 2024-11-29
units: allow -fFILE

mknos:
* BSD version allows FILE argument to -f option with and without a space
* Remove --file option because it is not a standard thing (DragonFlyBSD, NetBSD & OpenBSD don't have it)
* Previously the regex for -f option would allow bare '-' but now that is treated as an invalid option
* Also allow '--' to terminate options (found when testing against OpenBSD)


briandfoy (approved):

briandfoy:
changes: allow -fFILE (no space before argument)

PR #845, opened 2024-11-29
ed: disallow trailing '/' when setting filename

mknos:
* The usage "f abc/" is technically invalid because it would point to a directory
* Subsequent "e" or "w" commands fail because ed does not operate on directories
* GNU ed 1.20 added validation to prevent this usage; add it here because it would flag bad user input slightly earlier


briandfoy:
changes: disallow trailing '/' when setting filename

PR #846, opened 2024-11-29
id: exit 1 on failure

mknos:
* Avoid calling die() for better control of exit code
* The pod manual stated the exit code would be 1, but for the NoSuchUser case it wasn't

### DIFF
--- a/bin/id
+++ b/bin/id
@@ -20,13 +20,21 @@ License:
 #
 
 use strict;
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: exit 1 on failure (not die with 255)

PR #847, opened 2024-12-02
hexdump: retire -u option

mknos:
* Remove -u which does not exist in OpenBSD or util-linux version
* Add a version number to the script


briandfoy (approved):

briandfoy:
changes: remove non-standard -u (unbuffered) option

PR #849, opened 2024-12-02
ed: reject filename of '.' or '..'

mknos:
* Extending from commit 5f9377d8b67fbbb8dfbd745566bf3a9ea3299309, we know that filenames '.' and '..' are reserved for directories so make these names also raise an error in edFilename()
* The filename '...' is not reserved


briandfoy (approved):

briandfoy:
changes: reject filename of '.' or '..'

PR #850, opened 2024-12-02
hexdump: add more formats

mknos:
* Default output format switches to two-byte hex to match other versions, with canonical hex+ascii as -C option
* Add -b, -d, -o and -x format options as supported on OpenBSD
* Switch default output to filter duplicate lines, with -v option to disable this filtering
* Sync usage string and pod manual

### DIFF
--- a/bin/hexdump
+++ b/bin/hexdump
@@ -154,7 +154,7 @@ sub dump_dec2 {
     undef @chars;
 }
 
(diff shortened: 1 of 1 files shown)


briandfoy:
changes: add -b, -d, -o and -x format options

briandfoy (approved):

PR #852, opened 2024-12-03
od: remove global $len

mknos:
* perlcritic issued a warning that the variable $len in zeropad() was aliasing a global variable
* I found that $len doesn't need to be a global
* dump_file() and do_skip() save the read() return value into $len
* Beyond checking for a failed read, the value of $len was not needed


briandfoy (approved):

briandfoy:
I don't particularly care for perlcritic, but if you'd like to add a perlcriticrc that you like, we can start applying it. I figure that most programs will fail horribly. :)

PR #854, opened 2024-12-04
mkdir: compatible permissions

mknos:
* Doing a sync of function mod() from bin/chmod wasn't enough
* Call mod() without a file argument because symbolic permissions should assume a default value of 0777
* Now the mod() function diverges slightly from bin/chmod since $mode initialises to '777', not '000'
* The 0755 permission applies when no -m option is given, but 0777 is used as base permission when constructing a symbolic permission (which is calculated relative to a base permission)
* I tested some mkdir usages against OpenBSD...
* test1: mkdir d ---> permissions of 0755
* test2: mkdir -m a= d ---> no permissions (symbolic)
* test3: mkdir -m 0 d ---> no permissions (numeric)
* test4: mkdir -m u=rwx,g=rx ---> permissions of 0757 (o=7 because the default 0777 permission applies)
* test5: mkdir -m u=rwx,g=rx,o= ---> permissions of 0750 (because 'o' is explicitly updated

### DIFF
--- a/bin/mkdir
+++ b/bin/mkdir
@@ -17,7 +17,7 @@ use strict;
 use File::Basename qw(basename dirname);
 use Getopt::Std qw(getopts);
 
-our $VERSION = '1.3';
+our $VERSION = '1.4';
 my $Program = basename($0);
 
 $SIG{__WARN__} = sub { warn "$Program: @_\n" };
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: default permissions are 0777 unless you specify them

PR #855, opened 2024-12-04
glob: normal vs debug output

mknos:
* When enabling debug environment variable, print debug output to STDERR
* Currently the debug output interferes with the regular items printed to STDOUT
```
%export DEBUG_FASTGLOB=1
%for f in $( perl glob 'x*' ); do echo "F=$f"; done
regexp is x.*
component re is qr(\Ax.*\Z)
F=x.bc
F=x.pl
F=xaaa
F=xaab
F=xaac
F=xargs
F=xx
F=xxx
F=xxyy
```

### DIFF
--- a/bin/glob
+++ b/bin/glob
@@ -249,7 +249,7 @@ sub match_glob($) {
     }
 
     # debugging
-    print "regexp is $_\n" if ($verbose);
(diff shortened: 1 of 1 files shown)


briandfoy:
changes: debug info goes to STDERR instead of STDOUT

PR #856, opened 2024-12-04
tail: portable -f DIR

mknos:
* tail -f flag changes how all the file arguments are handled, i.e. tail_f() function is called
* When passing directory arguments to tail -f, ```@dirlist``` contains all the directory args
* get_existing_files() is called for each directory
* Switch get_existing_files() to use catfile() to join directory name with filenames from readdir()---this removes the assumption that the OS uses '/' as a directory separator

### DIFF
--- a/bin/tail
+++ b/bin/tail
@@ -29,6 +29,7 @@ $| = 1;
 use strict;
 
 use File::Basename qw(basename);
+use File::Spec;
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: use File::Spec to construct paths portably

PR #857, opened 2024-12-04
bc: convert trig functions to builtin

mknos:
* When running "bc -l", the return values for a(), c() and s() didn't seem right
* Instead of parsing these functions as bc code, I propose defining these as builtin perl functions (this is slightly easier to work with)
* Import Math::Trig for the benefit of atan()
* When running in bignum mode, s() function could hang; adapt a workaround of converting temporary argument to LittleNum so it can be passed to atan() etc [1]
* The mathlib functions e(), j() and l() are still defined as bc code but could be converted to builtin later

1. https://raw.githubusercontent.com/caldwell/pc/master/pc

### DIFF
--- a/bin/bc
+++ b/bin/bc
@@ -783,6 +783,8 @@ repository:
 
 use strict;
 
+use Math::Trig;
+
 # The symbol table : the keys are the identifiers, the value is in the
 # "var" field if it is a variable, in the "func" field if it is a
 # function.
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: use Math::Trig (released with 5.004) for trig

PR #858, opened 2024-12-05
hexdump: warn for directories

mknos:
* Avoid fatal error for directory arguments
* Print a warning for a directory, then proceed to next argument
* Use case tested: hexdump -C . .. file.txt


briandfoy (approved):

briandfoy:
changes: don't die for directory argument; warn and move to next argument

PR #860, opened 2024-12-05
unshar: print usage string

mknos:
* Instead of ignoring unknown options, show the user which options are expected
* User might type -e option, expecting it to be there because GNU sharutils has it


briandfoy (approved):

briandfoy:
changes: change to getopts

PR #861, opened 2024-12-05
bc: unhook mathlib

mknos:
* Convert the functions e(), j() and l() to builtin
* j() is actually the standard jn() function


briandfoy (approved):
Nice

briandfoy:
changes: change for GNU mathlib to internal perl

PR #862, opened 2024-12-06
glob: calculate $dirsep dynamically

mknos:
* Add a function to find an appropriate $dirsep value based on catfile()
* Now it won't be necessary to reconfigure $dirsep manually if the program is used on DOS/Windows
* No change of behaviour on my Linux system
* test1: perl glob 'a*' ---> current dir entries starting with 'a'
* test2: perl glob 'dir/*a*' entries in dir/ containing letter 'a'

### DIFF
--- a/bin/glob
+++ b/bin/glob
@@ -109,7 +109,6 @@ When invoking B<glob> as a function from the C<FastGlob> module, There
 are several module-local variables that can be set for alternate
 environments, they are listed below with their (UNIX-ish) defaults.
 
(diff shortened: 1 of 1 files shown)


bin/glob
```diff
@@ -267,6 +267,12 @@ sub match_glob {
     return sort(@res);
 }
 
+sub getdirsep {
```
briandfoy, line 270: This is an interesting way to do that, but is it better than the lookup table that File::Spec is using to decide which submodule to dispatch to?

briandfoy (approved):

PR #863, opened 2024-12-06
ed: standard exit code

mknos:
* When running an ed script the exit code is important
* If any command in the script fails, edWarn() will be called and $Error will be set; this should translate to a failure exit code
* Tested against GNU version  (both tests show successful commands after the failed command)
```
#test1: q command after invalid address error
470
*H
*100p
?
invalid address
*10p
	/* this? */
*q
%echo $?
1

#test2: wq command after invalid address error
470
*H
*100p
?
invalid address
*1,2wq new.txt
39
%echo $?
1
```

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -113,7 +113,7 @@ my $NO_QUESTIONS_MODE = 0;
 my $PRINT_NUM = 1;
 my $PRINT_BIN = 2;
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: always exit with 1 for failure

PR #864, opened 2024-12-06
tail: fix function params

mknos:
* When removing function prototypes I noticed two issues causing a fatal error thanks to strict.pm...
* The first 2 args of tail_f() are supposed to be array-ref
* The first 2 args of get_existing_files() are supposed to be array-ref
* With this patch, "perl tail -f regular.txt" doesn't die, and follows changes to the file as expected
* I guess with the old prototypes the parameters were being coerced, but that makes the code harder to audit
* Move one statement so it appears after "use strict"

### DIFF
--- a/bin/tail
+++ b/bin/tail
@@ -24,8 +24,6 @@ License: perl
 # 99/03/02 : first version
 #
 
-$| = 1;
-
 use strict;
 
 use File::Basename qw(basename);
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: some fixups to make strict happy

PR #865, opened 2024-12-06
bc: minarg checks for builtin functions

mknos:
* When previously expanding the number of builtin functions, I didn't notice that bi_sqrt() would incorrectly call sqrt(undef) if I type sqrt() at bc prompt
* Other versions of bc raise an error for this, and it's a good idea to flag bad function usage
* This patch adds min-args check to all the builtin functions; all take a single arg except j() which takes two
* When I pass too many args to c() I already see an error

### DIFF
--- a/bin/bc
+++ b/bin/bc
@@ -2162,6 +2162,7 @@ sub bi_length
   my $stack = shift;
 
   $_ = pop @$stack;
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: catch situations where there are missing arguments to functions

PR #867, opened 2024-12-09
glob: don't raise error for unknown user with '~'

mknos:
* A tilde can optionally be followed by a username
* If a provided username is not found, treat the input as no-match instead of raising an error
* This is consistent with Linux/bash, OpenBSD/ksh and the perl builtin glob() function
* If '~' appears without a username, it is substituted for the current user's home directory
* Remove some '&' sigils from &xzy() function calls---glob() now gets called with fully qualified name
```
%echo ~mail/a # valid user
/var/mail/a
%echo ~mayel/a # not found
~mayel/a
%echo ~/a
/home/pi/a
```

### DIFF
--- a/bin/glob
+++ b/bin/glob
@@ -180,7 +180,7 @@ use File::Spec;
 
 use vars qw($VERSION @ISA @EXPORT @EXPORT_OK);
 
-$VERSION = 1.2_06;
+$VERSION = 1.2_07;
 @ISA = qw(Exporter);
 @EXPORT = qw(&glob);
 @EXPORT_OK = qw(dirsep rootpat curdir parentdir hidedotfiles);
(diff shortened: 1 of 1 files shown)


bin/glob
```diff
@@ -180,7 +180,7 @@ use File::Spec;
 
 use vars qw($VERSION @ISA @EXPORT @EXPORT_OK);
 
-$VERSION = 1.2_06;
+$VERSION = 1.2_07;
```
briandfoy, line 183: We can probably make this a non-trial version unless you have a reason you'd like to keep it as such.

briandfoy (approved):

briandfoy:
changes: with ~someuser, if that user is not found, treat it as a non-match instead of an error

PR #869, opened 2024-12-09
ed: add undo feature

mknos:
* The undo command (u) was listed in a TODO comment
* There's only 1 level of undo, i.e undo will undo itself
* This implementation which assumes /tmp has enough space to store a copy of the buffer
* Write a new undo buffer if a modification was detected
* According to other versions, the current line pointer should be updated too, but I want to handle this later
* Remove some ancient comments; perl coolness *is* a guiding principle
* Document undo in POD

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -33,22 +33,10 @@ License: gpl
 #
 # Irony:  This was, of course, edited originaly with emacs...
 #
-# Commentary:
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: add undo (u) command

PR #870, opened 2024-12-09
mail: -f usage forbids extra args

mknos:
* If mail command is invoked with arguments it assumes those are addresses to send mail to, but running "mail -f mbox" is not compatible
* Raise an error and print usage string, as done in OpenBSD version
* Also use catfile() when constructing a default value for mailbox path


briandfoy (approved):

briandfoy:
changes: -f usage forbids extra args

PR #871, opened 2024-12-09
ed: update current line in undo

mknos:
* GNU and OpenBSD versions restore line number at same time as restoring editor buffer in undo operation
* This is needed because $CurrentLineNum might not point to a valid address anymore
* For example, run "$a" command to append lines to end of buffer (currentLine becomes new maxLine), then run "u" to undo the append (buffer shrinks, maxLine reverts)


briandfoy (approved):

briandfoy:
changes: update current line in und

PR #874, opened 2024-12-16
tail: less confusing signals for -f flag

mknos:
* GNU version terminates "tail -f" instance on ctrl-c
* Allowing ctrl-c to terminate program will do what the user expects
* Steal the existing SIGQUIT handler for the purpose of showing file report
* Update pod manual; only SIGQUIT needs to be mentioned now

### DIFF
--- a/bin/tail
+++ b/bin/tail
@@ -278,8 +278,7 @@ sub tail_f
     @filelist = @{ $files };
     @dirlist = @{ $dirs };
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: the INT signal now terminates the program, while the QUIT signal shows the files being watched

PR #876, opened 2024-12-17
touch: -t0 is invalid

mknos:
* Truth check for option -t argument prevented validation in parse_time() from running
* "touch -t 0" isn't valid because a timespec requires >1 character
* Style: remove "next" statement on utime() failure because there is no subsequent code to skip


briandfoy (approved):

briandfoy:
changes: check -t argument for definedness, not truth

PR #878, opened 2024-12-17
pr: -a flag not recognised

mknos:
* A typo in the options handling code prevented $roundrobin from being set
* This patch allows the round-robin logic to be verified (I haven't investigated the output for -a yet)


briandfoy (approved):

briandfoy:
changes: correctly detect -a (previously ignored due to bug)

PR #880, opened 2024-12-18
sort: -o0 is valid

mknos:
* The -o option writes sorted data to a file
* A truth check in init code was preventing output file 0 from being used
* The value was set back to empty string, which is interpreted later in _merge_files() as meaning stdin


briandfoy (approved):

briandfoy:
changes: the -o switch can accept 0 (zero) as a filename (check defined, not truth)

PR #881, opened 2024-12-18
lock: re-enable term echo

mknos:
* When testing lock with a 1 minute timeout, my terminal text was invisible after lock times out and exits
* In the alarm handler, be careful to enable echo before the program finishes


briandfoy (approved):

briandfoy:
changes: re-enable term echo after lockout

PR #883, opened 2024-12-18
what: bypass readline for directory argument

mknos:
* Directory arguments are treated as a file that doesn't contain any matches
* Add explicit directory handling code to avoid reading a line from a directory, which would always fail
* Tested against OpenBSD
```
%perl what .. /bin/bash not-exist
..:
/bin/bash:
	Bash version 5.1.4(1) release GNU
Unable to open 'not-exist': No such file or directory
%echo $?  # bash contained a match
0
```

### DIFF
--- a/bin/what
+++ b/bin/what
@@ -22,6 +22,10 @@ getopts('s', \%opt) or usage();
 
 for my $file (@ARGV)
     {
+        if (-d $file) {
+            print $file, ":\n";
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: don't try to read lines from directories - directories are "files" that never match

PR #887, opened 2025-01-01
mkdir: reduce chmod() code

mknos:
* Improve code clarity by directly processing numeric file mode, instead of converting mode to a string and then back to a number
* The warning printed when chmod() failed incorrectly stated that stat() had failed


briandfoy (approved):

briandfoy:
Thanks for taking a break over the holidays :)

PR #888, opened 2025-01-01
touch: options -r and -t are incompatible

mknos:
* Interpret options -r and -t as mutually exclusive
* The standards document provides a usage string which clarifies this [1]
* Previously if -r and -t were used together, one would be silently ignored
* Ignoring one option is confusing because it's unclear to the user which time value was actually used
* Style change: quote some hash keys to satisfy vim syntax highlighting ("m}" was seen as the start of a regex)

1. https://pubs.opengroup.org/onlinepubs/9699919799/utilities/touch.html

### DIFF
--- a/bin/touch
+++ b/bin/touch
@@ -23,7 +23,7 @@ use constant EX_FAILURE => 1;
 
 my $Program = basename($0);
 
-my ($VERSION) = '1.3';
+my ($VERSION) = '1.4';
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: -r and -t are mutually exclusive and together will throw an error

PR #891, opened 2025-01-01
primes: raise error for start>end

mknos:
* If run with 2 arguments, primes are displayed for a range from Start to End
* Normally an error was raised if End was smaller than Start, but the value of 0 was resulting in a default value unexpectedly
* Convert maximum to a constant so it could be changed easier to 64 bits (no dynamic size calculation is being done but at least the user can patch the one constant)
* Original test: perl primes 10 0

### DIFF
--- a/bin/primes
+++ b/bin/primes
@@ -15,12 +15,15 @@ License: perl
 
 use strict;
 #use integer; # faster, but cuts the maxint down
+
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: disallow 0 as an upper bound (inadvertently chooses the default)

PR #893, opened 2025-01-02
mkdir: suppress multiple errors for non-directory argument

mknos:
* Previously if running "mkdir -p" and the 1st part of an argument is not a directory, the program would incorrectly call mkdir() on the intermediate paths under it
* If I have a regular file 0, "mkdir -p 0/1/2" would show 3 errors because of mkdir("0/1") and mkdir("0/1/2")
* Restructure the program so ARGV can remain as a list of directories (now we don't need the explanation comment for the old confusing code)
* For -p flag, split directory into parts, calling a create_dir() function for each part
* split_dir() handles the case where the 1st part is not a directory, and drops the sub-directory parts

```
%touch 0
%perl mkdir -p 0/1/2 00 aa/bb/cc # only one error for 0/1/2 argument
mkdir: cannot create directory '0': File exists

%find 00 aa # everything was created except for 0/...
00
aa
aa/bb
aa/bb/cc
```

### DIFF
--- a/bin/mkdir
+++ b/bin/mkdir
@@ -14,7 +14,8 @@ License: perl
 
 use strict;
 
-use File::Basename qw(basename dirname);
+use File::Basename qw(basename);
+use File::Spec;
 use Getopt::Std qw(getopts);
 
 our $VERSION = '1.6';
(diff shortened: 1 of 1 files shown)


briandfoy (changes requested):
This isn't a request for a change really, but since File::Path and File::Spec are core modules, is there a chance to use those here? We can still merge this change and think about that later, too.

mknos:
Thanks for the suggestion. I added a new commit which rewrites split_dir() using File::Spec.
I found another test case where subdir creation should bail out: permission denied error when creating parent dir.
create_dir() now returns true on success so the subdir creation can stop on error.
This means split_dir() has no need to do any file tests.
Handle special case where File::Spec adds a leading '' list element for absolute paths.
```
%perl mkdir -p 0/1/2 /0/1/2 aa/bb/cc 00 # single error for 0/1/2 and /0/1/2 respectively
mkdir: cannot create directory '0': File exists
mkdir: cannot create directory '/0': Permission denied

%find aa 00
aa
aa/bb
aa/bb/cc
00
```

briandfoy (approved):

briandfoy:
changes: suppress multiple errors for non-directory argument

PR #894, opened 2025-01-02
clear: handling OSPEED is not needed

mknos:
* The OSPEED option being passed to Tgetent() is not required for clearing the screen
* Win32 perl might not have a working Term::Cap, so move the Tgetent() call into the eval just in case
* Tested on Linux and OpenBSD

### DIFF
--- a/bin/clear
+++ b/bin/clear
@@ -11,21 +11,13 @@ License: perl
 
 =cut
 
-
-use Term::Cap;
 use strict;
 
-my $OSPEED = 9600;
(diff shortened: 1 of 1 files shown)


briandfoy:
I see this, and I'll test it on Windows then merge if there are no problems.

briandfoy:
This works for me on Windows 10 using:

* ComEmu
* PowerShell
* Command Prompt

I think all of these are using the same underlying code to manage the window.

PR #896, opened 2025-01-03
fish: retire fisher_yates_shuffle()

mknos:
* Reduce code that needs to be maintained by using the shuffle() function from List::Util
* Small declarations change: explicitly import getopts() function

### DIFF
--- a/bin/fish
+++ b/bin/fish
@@ -11,9 +11,11 @@ License: perl
 
 =cut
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: use List::Util instead of implementing our own shuffle

PR #897, opened 2025-01-03
grep: retire GREP_OPTIONS

mknos:
* I discovered that GNU grep 3.6 on my Linux system completely ignores GREP_OPTIONS
* I did a little reading and the feature had been removed from GNU grep (probably for security reasons)
* Original BSD grep and current OpenBSD grep don't have GREP_OPTIONS, so remove it (it's not a standard thing anyway)


briandfoy (approved):

briandfoy:
changes: remove GREP_OPTIONS environment variable, which GNU grep removed (security)

PR #898, opened 2025-01-03
paste: factor loop

mknos:
* The input loop was written as a forever loop with an eof check at the top of loop body
* Convert the eof check to a function so the loop can be written as a "while"
* Also add --version as done in other scripts
* Regression test: perl paste /etc/motd /etc/motd


briandfoy (approved):

briandfoy:
changes: add --version; light refactoring

PR #901, opened 2025-01-03
mkfifo: remove a global variable

mknos:
* Avoid calculating the default mode in the case where a numeric mode is supplied to -m (get_mode() just returns oct($opt_m) here
* Defer default mode calculation to a function which is used for 1) no -m option and 2) symbolic permission passed to -m
* Passing a value to umask() is not necessary because we only want to get the current value, and the mkfifo() call uses an explicit mode param

### DIFF
--- a/bin/mkfifo
+++ b/bin/mkfifo
@@ -28,8 +28,9 @@ my $Program = basename($0);
 
 getopts('m:') and @ARGV or usage();
 
-my $default_mode = 0666;
-$default_mode &= ~(umask 0);
+sub default_mode {
+	return 0666 & ~(umask);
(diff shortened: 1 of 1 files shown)


bin/mkfifo
```diff
@@ -28,8 +28,9 @@ my $Program = basename($0);
 
 getopts('m:') and @ARGV or usage();
 
-my $default_mode = 0666;
-$default_mode &= ~(umask 0);
+sub default_mode {
```
briandfoy, line 31: I like this sort of change.

briandfoy (approved):

PR #903, opened 2025-01-04
tac: exit(1) on error

mknos:
* When processing multiple file arguments, and one file fails to open, exit(1) after processing all arguments to indicate something went wrong
* Found when testing against GNU tac

```
%echo \\0/ > exist && perl tac . .. notexist exist 
tac: '.' is a directory
tac: '..' is a directory
tac: failed to open 'notexist': No such file or directory
\0/
%echo $?
1
```


briandfoy (approved):

briandfoy:
changes: exit 1 on error

PR #904, opened 2025-01-04
patch: fix external ed invocation

mknos:
* This version of patch first attempts to process an ed diff by running an external ed command, then falls back to processing the ed commands directly
* I noticed the external ed code was always failing but I didn't debug it until now
* Problem1: SIGCHLD happens normally when ed exits but it was being treated as an error (removing signal handlers fixes it)
* Problem2: write command sent to ed was not terminated by a newline, so ed would never run the command
* Problem3: after writing to o_file, the offset of filehandle $out was at byte 26, so the test line comparison wasn't correct
* With this patch I can process a simple ed diff on Linux without the code reaching PLAN_J

```
%cat ab.diff # change a.c into b.c
9,14d
%cp a.c acopy.c
%perl patch -e acopy.c ab.diff
Hmm...  Looks like an ed diff to me...
Patching file acopy.c using Plan A...
Hunk #1 succeeded at 1.
done
%cmp acopy.c b.c
%cmp acopy.c a.c
acopy.c a.c differ: byte 84, line 9
```

### DIFF
--- a/bin/patch
+++ b/bin/patch
@@ -912,8 +912,7 @@ sub apply {
 
     # Pipe to ed.
     eval {
-        my $cmd =  'ed -s ' . $self->{'i_file'};
-        open ED, '|-', $cmd or die "Couldn't fork ed: $!";
+        open ED, '|-', 'ed', '-s', $self->{'i_file'} or die "Couldn't fork ed: $!";
         print ED map @$_, @cmd               or die "Couldn't print ed: $!";
         print ED "wq $self->{o_file}\n"     or die "Couldn't print ed: $!";
         close ED                             or die "Couldn't close ed: $?";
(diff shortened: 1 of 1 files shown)


bin/patch
```diff
@@ -912,16 +912,17 @@ sub apply {
 
     # Pipe to ed.
     eval {
-        local $SIG{PIPE} = sub { die 'Pipe broke...' };
-        local $SIG{CHLD} = sub { die 'Bad child...' };
-        open ED, '|-', qw(ed - -s), $self->{i_file} or die "Couldn't fork ed: $!";
+        my $cmd =  'ed -s ' . $self->{'i_file'};
+        open ED, '|-', $cmd or die "Couldn't fork ed: $!";
```
briandfoy, line 916: Was there a problem with the list form of the pipe? That should be safer.

briandfoy (changes requested):

bin/patch
mknos: Well spotted. I added a commit to convert the ed arguments back to a list and repeated the original test.

briandfoy (approved):

PR #905, opened 2025-01-06
tsort: '-' argument is not special

mknos:
* Standards document does not require that a '-' argument is interpreted as stdin [1]
* Follow BSD semantics and treat '-' as a regular file argument (tested against OpenBSD)
* Previous special meaning of '-' was not documented in the pod manual; add explicit text to explain when stdin is used 

1. https://pubs.opengroup.org/onlinepubs/9699919799/utilities/tsort.html

### DIFF
--- a/bin/tsort
+++ b/bin/tsort
@@ -30,12 +30,8 @@ if (@ARGV) {
     warn "$Program: extra operand '$ARGV[0]'\n";
     usage();
 }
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: don't treat - as a special filename for stdin

PR #908, opened 2025-01-06
date: catch invalid date arguments

mknos:
* Extra format string arguments are ambiguous and should result in an error
* select_format() is meant to select only one format; the use of grep() as a filter prevented it from catching invalid arguments
* Arguments which are not a valid format string should result in an error
* Found when testing against GNU date
* test1: perl date    # default in select_format()
* test2: perl date yo+%a   # not a valid date format
* test3: perl date +%a +%a    # too many formats
* test4: perl date +%a%a   # still works

### DIFF
--- a/bin/date
+++ b/bin/date
@@ -16,7 +16,7 @@ License: artistic2
 
 use POSIX;
 
-my $VERSION = '1.0.4';
+my $VERSION = '1.0.5';
 
 # set this if we do anything to select a timezone abbreviation, then
 # prefer this if it is defined.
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: all formats must start with + and there can only be one

briandfoy:
I had to back out this change because it broke `date -u`, and maybe other things.

bin/date
```diff
@@ -80,8 +80,19 @@ sub run {
 	}
 
 sub select_format {
-	my( $supplied ) = grep { /\A\+/ } @_;
-	return $1 if $supplied =~ /\A\+(.+)/;
+	my $supplied;
```
briandfoy, line 83: I ended up backing out this change. I should have paid attention to the test failing, but there was another test failing so I just ignored both. That's dumb.

A few things to note here:

* don't exit from utility subs. In the fix I die and wrap the sub in an eval. That gives the main code a chance to ignore it. Note a huge deal here because it's all private code, but in general it's about separation of concerns.
* we can't stop if there is an option that doesn't start with `+` because there are other options possible, and maybe we need better argument processing so nothing else is left.

PR #909, opened 2025-01-06
uniq: pod fix

mknos:
When I run pod2text on uniq I see "compar- isons" printed on the same line. Allow the pod formatter to break lines where it sees fit, instead of manually hyphenating words.


briandfoy (approved):

briandfoy:
changes: undo some line wrapping hyphenation in docs. Let the formatter figure it out

PR #910, opened 2025-01-07
fmt: avoid space before newline

mknos:
* In fmt_line(), the ```@out``` list describes one line of text
* Each word token keeps a record of trailing spaces after the word (i.e. "suffix" number in the code)
* Trailing space should be ignored for the final word on the line because it gets terminated by a newline

### DIFF
--- a/bin/fmt
+++ b/bin/fmt
@@ -147,6 +147,7 @@ sub fmt_line {
     my $s = ' ' x $out[0]->{'indent'};
     print $s;
 
+    $out[-1]->{'suffix'} = 0;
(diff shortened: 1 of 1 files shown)


bin/fmt
```diff
@@ -147,6 +147,7 @@ sub fmt_line {
     my $s = ' ' x $out[0]->{'indent'};
     print $s;
 
+    $out[-1]->{'suffix'} = 0;
```
briandfoy, line 150: these seems like it will work for now.

briandfoy (approved):

PR #912, opened 2025-01-07
rm: simplify process_options()

mknos:
* is_recursive() is already smart enough to check both -R and -r, so manually setting ```$opts{R}``` is not required
* As a regression test I removed some subdirectories, first using -r, then -R

### DIFF
--- a/bin/rm
+++ b/bin/rm
@@ -196,8 +196,6 @@ sub process_options {
 		);
 	usage() unless $ret;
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):
hey, you finally touch something I did. :)

briandfoy:
I adjusted the test for this script since it looked at the internals to see what options were set. No big whoop.

PR #915, opened 2025-01-08
diff: simplify LCS_matrix()

mknos:
* Initialise $x[$i] to new listref [0] in one statement, instead of assigning list element 0 separately
* Calling standard max() function avoids temporary variables when compared items are different
* Write loops in idiomatic form with ranges
* This was intended as a readability change, but I observed that when generating diffs on a large text file the performance is slightly improved
```
%wc -l bc bc.new 
  2761 bc
  2768 bc.new
  5529 total
%time perl diff.old -u bc bc.new > /dev/null
real	0m11.003s
...
real	0m10.845s
...
real	0m10.965s

%time perl diff.new -u bc bc.new > /dev/null
real	0m9.470s
...
real	0m9.546s
...
real	0m9.329s
```

### DIFF
--- a/bin/diff
+++ b/bin/diff
@@ -609,6 +609,8 @@ package
 
 use strict;
 
+use List::Util qw(max);
+
 our $VERSION = '0.57';
 
 sub LCS_matrix {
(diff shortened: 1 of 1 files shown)


briandfoy (approved):
Nice.

PR #916, opened 2025-01-08
ed: fix append bug

mknos:
* It is valid usage to start ed with a filename that doesn't exist; ed is expected to start with an empty buffer and the default filename variable set
* Today I found that commands 0i and 0a were resulting in the first line of input being lost
* When debugging, I discovered the ```@lines``` list was empty when ed was started with a non-existent file
* A placeholder item $lines[0] is expected to be populated, but edEdit() was returning early because the file doesn't exist
* Fix this by initialising $lines[0] at start of program, i.e. before the 1st edEdit() call


briandfoy:
changes: initialize line placeholder so lines aren't lost after stating with an empty file

PR #919, opened 2025-01-17
ed: e! should not set filename

mknos:
* I hadn't tested this scenario before
* First I start ed with a file a.c
* The command "e !ifconfig" then replaces the buffer with the output of ifconfig
* Listing current filename with "f" command after this should still show "a.c" (this is how GNU and OpenBSD versions behave)
* The directory check should not happen for "e !..." case

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -105,7 +105,7 @@ my $NO_QUESTIONS_MODE = 0;
 my $PRINT_NUM = 1;
 my $PRINT_BIN = 2;
 
-our $VERSION = '0.25';
+our $VERSION = '0.26';
 
 my @ESC = (
     '\\000', '\\001', '\\002', '\\003', '\\004', '\\005', '\\006', '\\a',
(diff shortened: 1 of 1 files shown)


bin/ed
```diff
@@ -717,13 +723,13 @@ sub edEdit {
         return;
     }
 
-    if (-d $filename) {
-        warn "$filename: is a directory\n";
-        return E_READ;
-    }
-    if ($filename =~ s/\A\!//) {
+    if ($do_pipe) {
         return unless (open $fh, "$filename |"); # no error
```
briandfoy, line 727: I think this `open` needs some attention. First, it shouldn't use the magic open with the `|` and the filename in the same string. Second, should there be a warning when the open fails, as there is in the `-d $filename`  case right after it?

Maybe that's too much surgery for this issue and we could push it off into a separate issue.

mknos: Printing $! to stderr, as done by GNU ed, would be desirable at least in case the open() fails.
With the string argument to open(), I can make a patch that splits the input into a list with quotewords().
But I'd like to submit a patch for another issue related to saving the filename before this.

PR #921, opened 2025-01-18
ed: 0k is not OK

mknos:
* Avoid saving address 0 as a "marked address"
* When edMark() was first written I didn't pick up that this usage was not allowed in other versions of ed
* test1: 1kx --> save address 1 as mark "k", valid if the buffer is not empty
* test2: 0ky --> save address 0, invalid
* test3: kz --> save current address as mark "z", valid if the buffer is not empty


briandfoy (approved):

briandfoy:
changes: Avoid saving address 0 as a "marked address"

PR #923, opened 2025-01-19
ed: consistent filename validation

mknos:
* Extend the filename validation from the f command to the commands e and r, which also take a filename argument
* Now the code can fail slightly earlier, before open_file_ro() is called


briandfoy (approved):
Very nice refactoring

PR #924, opened 2025-01-20
ed: w! command should not update saved filename

mknos:
* I found that the "w !CMD" case had the same problem as "r !CMD"; avoid updating $RememberedFilename when writing lines from buffer to an external command thru a pipe
* Now init_pipe() takes an extra WriteFlag argument so edWrite() can use it too
* For consistency add illegal_file() check in edWrite() before output file is opened
* Also add quotewords() in edPipe() to allow system() to be passed an argument list
* test1: ```1,2w !rev``` --> pipe buffer lines 1-2 into rev command
* test2: ```$w ..``` --> illegal_file() == TRUE
* test3: ```!cc secret.c``` --> build secret project file

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -401,7 +401,10 @@ sub edPipe {
     return E_ADDREXT if defined $adrs[0];
 
     if (defined $args[0]) {
-        my $rc = system $args[0];
+        return E_SUFFBAD if $args[0] =~ m/\0/;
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: w! does nut update the saved filename

PR #927, opened 2025-01-21
od: support 1-byte hex format

mknos:
* od is expected to support a -t option for specifying output formats that don't exist as a single letter option
* 2- and 4-byte hex output is possible without -t, but "-t x1" is the normal way to request 1-byte hex
* Other format specifiers can be added to -t later, but for now add x1, x2 and x4
* Update pod and usage string


briandfoy:
changes: support 1-byte hex format

briandfoy (approved):

PR #928, opened 2025-01-22
od: support more output formats with -t

mknos:
* Add other integer format options (u1,u2,u4,d1,d2,d4,o1,o2,o4), as well as the aliases for options -a and -c
* 8-byte integer formats could be added in future (these are supported by GNU version)
* Make more of an effort to align output for -a and -c modes

### DIFF
--- a/bin/od
+++ b/bin/od
@@ -53,14 +53,14 @@ my %charname = (
     5  => 'enq',
     6  => 'ack',
     7  => 'bel',
-    8  => 'bs',
-    9  => 'ht',
-    10 => 'nl',
-    11 => 'vt',
-    12 => 'ff',
-    13 => 'cr',
-    14 => 'so',
-    15 => 'si',
+    8  => ' bs',
+    9  => ' ht',
+    10 => ' nl',
+    11 => ' vt',
+    12 => ' ff',
+    13 => ' cr',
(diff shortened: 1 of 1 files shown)


bin/od
```diff
@@ -53,14 +53,14 @@ my %charname = (
...
     7  => 'bel',
-    8  => 'bs',
-    9  => 'ht',
-    10 => 'nl',
-    11 => 'vt',
-    12 => 'ff',
-    13 => 'cr',
-    14 => 'so',
-    15 => 'si',
+    8  => ' bs',
```
briandfoy, line 56: hey, just want to check that these spaces are intentional. If they are, can you leave a comment next to the first instance (or somewhere close) that the spaces belong there? I can see some future editor not understanding this change and blindly removing the spaces.

briandfoy (changes requested):

bin/od
mknos: I followed your suggestion and put a comment above the ascii names table. I had removed a sprintf call from char7bit().

briandfoy (approved):

briandfoy:
changes: add other integer format options (u1,u2,u4,d1,d2,d4,o1,o2,o4), as well as the aliases for options -a and -c

PR #930, opened 2025-01-23
od: make float formats accessible with -t

mknos:
* Support arguments "-t f4" and "-t f8", which are already exposed as options -f and -F respectively
* In future, od can support multiple formats at once, e.g. "-t f4f8", but for now only one format is allowed


briandfoy (approved):

briandfoy:
changes: support arguments "-t f4" and "-t f8", which are already exposed as options -f and -F respectively

PR #931, opened 2025-01-24
ed: honour -s flag for w & e commands

mknos:
* In BSD ed, the -s flag is for running an ed script
* Modified buffers don't prevent the editor from quitting or editing a new file in this mode
* GNU ed doesn't seem to follow this, but to me the GNU interpretation seems inconsistent with the meaning of -s


briandfoy (approved):

briandfoy:
changes: honour -s flag for w & e commands

PR #932, opened 2025-01-27
ed: buffer is dirty unless all lines were written

mknos:
* I found a case where ed allowed unsaved changes to be lost
* The default behaviour of "w" command is to write all lines from the editor buffer
* In the case of 1w or 1,2w command, only the selected lines are written and there may be unsaved changes on other lines, so don't unset the "dirty" flag
* The flag should be unmodified also in the case of "w !cat", where all buffer lines are written to a pipe
* This patch makes ed consistent with BSD and GNU versions

```
%perl ed -p 'edmond>' a.s # desired behaviour
839
edmond>10d 
edmond>1w tmp.txt
13
edmond>q
?
edmond>h
buffer modified
edmond>Q
```

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -679,8 +679,12 @@ sub edWrite {
         warn "$filename: $!\n";
         return E_CLOSE;
     }
-
-    $NeedToSave = $UserHasBeenWarned = 0;
+    unless ($do_pipe) {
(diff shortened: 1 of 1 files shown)


briandfoy:
changes: buffer is dirty unless all lines were written

briandfoy (approved):

PR #933, opened 2025-01-27
moo: distinct-digits rule was not followed

mknos:
* In debugging I discovered the numbers 8182 and 4594 were being treated as valid
* The unix version of moo selects a number where each digit occurs only once in the number [1]
* A guess containing non-unique digits is not a valid guess [2]
* Add a function to check for duplicate digits in guess
* Rewrite secret number generator with shuffle() to guarantee unique digits

1. http://squoze.net/UNIX/v6man/man6/moo.pdf
2. https://github.com/hansklav/MOO

### DIFF
--- a/bin/moo
+++ b/bin/moo
@@ -11,13 +11,24 @@ License: perl
 
 =cut
 
-
 use strict;
 
-my ($VERSION) = '1.2';
+use List::Util qw(shuffle);
+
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: enforce the distinct digits rule

PR #935, opened 2025-01-29
maze: reject fractional width/height

mknos:
* The size of the maze is counted in whole numbers; fractions don't maze sense
* get_number() is used for input validation
* Make get_number() reject the input if the number contains non-digits
* Numbers don't allow a '+' prefix as in some commands
* If EOF was seen when prompting for a number, terminate the program
* test1: ```perl maze 11 11``` --> valid, no prompt
* test2: ```perl maze 12.34 12.6666666666666666666``` --> invalid, will prompt for both width & height

### DIFF
--- a/bin/maze
+++ b/bin/maze
@@ -51,14 +51,29 @@ usage() if @ARGV;
 
 sub get_number {
   my ($prompt, $value) = @_;
+  my $regmatch = 0;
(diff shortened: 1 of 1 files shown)


briandfoy:
changes: reject fractional width/height

briandfoy (approved):

PR #936, opened 2025-01-29
sum: add missing sha224 algorithm

mknos:
* sha224 was missing (provided by perl in Digest::SHA)
* Take advantage of other Digest algorithms which may already be installed from CPAN (but don't require them to be installed)
* Add SHA3 functions as well as JH, Whirlpool, Haval256, BLAKE, MD2 and MD4
* List optional algorithms in usage string and POD (don't attempt to detect which are installed prior to use)
* NB: Digest::MD6 fails to build and I sent an email to the maintainer (for now don't include MD6)
* In future the different algorithm functions could be merged into a common function since they mostly use the Digest.pm API

### DIFF
--- a/bin/sum
+++ b/bin/sum
@@ -184,7 +184,7 @@ sub do_md4 {
 		require Digest::MD4;
 		1;
 	} or do {
-		die "The md5 algorithm is not available on your system\n";
+		die "The md4 algorithm is not available on your system\n";
 	};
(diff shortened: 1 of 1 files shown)


briandfoy:
changes: add missing sha224 algorithm

briandfoy (approved):

PR #937, opened 2025-01-30
od: add -h alias for -x

mknos:
* GNU and OpenBSD versions support either -h or -x for 2-byte hex, and -H/-X for 4-byte hex
* This version had -X, -x and -H but not -h


briandfoy:
changes: add -h alias for -x

PR #939, opened 2025-01-31
od: add -D option

mknos:
* GNU and OpenBSD versions support -d for 2-byte unsigned decimal and -D for 4-byte unsigned decimal
* This versions was missing -D; it's a shortcut option for -t u4
* The OpenBSD manual incorrectly states that -D is an octal format[1]
* The OpenBSD source confirms the format is unsigned decimal[2]

1. http://man.openbsd.org/od
2. https://cvsweb.openbsd.org/cgi-bin/cvsweb/src/usr.bin/hexdump/odsyntax.c?annotate=1.28 (L127)

### DIFF
--- a/bin/od
+++ b/bin/od
@@ -23,11 +23,11 @@ use constant EX_FAILURE => 1;
 use constant LINESZ => 16;
 use constant PRINTMAX => 126;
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: add -D option as alias for -t u4 (four digit decimals)

PR #940, opened 2025-02-05
diff: add -s option

mknos:
* GNU and OpenBSD versions support the -s flag, which is similar to -q
* -q shows a "files differ" messages instead of showing the diff
* -s explicitly tells the user when the files are the same, instead of silently exiting
* Improve how -q is documented; it doesn't print anything for identical files
* -q and -s are not mutually exclusive
* While here, remove manual assignments of $opt_u and $opt_c which are not needed (the subsequent code refers to $Context_Lines and not these variables)
```
%perl diff -qs a.c a.c
Files a.c and a.c are identical
%perl diff -qs a.c a.s
Files a.c and a.s differ
```

### DIFF
--- a/bin/diff
+++ b/bin/diff
@@ -44,7 +44,7 @@ use constant EX_SUCCESS   => 0;
 use constant EX_DIFFERENT => 1;
 use constant EX_FAILURE   => 2;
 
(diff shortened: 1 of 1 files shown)


briandfoy:
changes: add -s to show when files are the same instead of silently exiting

PR #941, opened 2025-02-06
patch: add paths to error messages

mknos:
* Improve error messages if open() fails
* INFILE was listed instead of the original file path
* OUTFILE was listed instead of the path provided to -o, e.g. if I specify an invalid argument
* Style: switch filehandles to my-variables
* Style: add more quotes for hash keys

```
%perl diff env.c env2.c | perl tee env.diff
61d60
< 	setprogname(*argv);
79d77
< 		case '?':
%perl patch -o /G/G env.c env.diff 
Patching file env.c using Plan B...
Couldn't open OUTFILE: No such file or directory
Skipping patch...

% chmod 0 env.c
%perl patch -o NEW_ENV env.c env.diff
Patching file env.c using Plan B...
Couldn't open INFILE: Permission denied
Skipping patch...
```

### DIFF
--- a/bin/patch
+++ b/bin/patch
@@ -24,7 +24,7 @@ use constant EX_SUCCESS => 0;
 use constant EX_REJECTS => 1;
 use constant EX_FAILURE => 2;
 
-my $VERSION = '0.31';
+my $VERSION = '0.32';
 
 $|++;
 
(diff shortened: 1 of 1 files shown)


briandfoy:
nice change, thanks

PR #944, opened 2025-02-07
patch: throw() regression

mknos:
* I noticed a runtime error within Patch::Ed::apply() where the throw function can't be resolved
* This error didn't happen when I switch back to historical commit 8a1ef6abe04b811ec9040bd4810e39cd73ca0945
* The Patch::import() function had been removed, and this was forcibly putting throw into the caller's namespace
* None of the other subclasses of Patch call throw(), only Patch::Ed does
* The old import() approach was not the Right Thing so I don't want to revert to it
* Make throw() an oo method instead of a simple function; this allows perl to resolve it from the parent class
* Delete an incorrect comment that origfile comes from -o option (-o is output and origfile is input)

```
%perl patch -u a.c patch.ed
Not a unified diff!
Skipping patch...
Undefined subroutine &Patch::Ed::throw called at patch line 901, <$fh> line 208.
```

### DIFF
--- a/bin/patch
+++ b/bin/patch
@@ -24,7 +24,7 @@ use constant EX_SUCCESS => 0;
 use constant EX_REJECTS => 1;
 use constant EX_FAILURE => 2;
 
-my $VERSION = '0.33';
+my $VERSION = '0.34';
 
 $|++;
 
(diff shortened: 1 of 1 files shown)


briandfoy:
changes: fix the unresolvable throw() encountered when skipping a patch

PR #948, opened 2025-02-10
ed: compatible \Z and \z in substitute

mknos:
* Avoid special meaning of \Z and \z escapes when substituting text
* The special meaning of \n was already prevented in this way
* Add a missing call to get_terminated_line() in edSubstitute() that was found when testing this (the result of the s command is conditionally printed)
* It is still possible to bind a substitution to the end of a line using ```s/$/text/```
* Tested against GNU ed

### DIFF
--- a/bin/ed
+++ b/bin/ed
@@ -548,6 +548,9 @@ sub edSubstitute {
     }
 
     $args[0] =~ s/\\n/n/g; # newline is not permitted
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: treat \z and \Z in substitutions as literal z and Z

PR #949, opened 2025-02-12
morse: fold case bug

mknos:
* Sometimes letter case was being folded and sometimes it was not
* When debugging, txt2mors() handled the folding but txt2word() was missing this logic; both of these functions lookup values in %chartab

```
%perl  morse aia
dit daw 
dit dit 
dit daw 
%perl  morse AIA
%perl  morse -s aia
.-
..
.-
%perl  morse -s AIA
.-
..
.-
```

### DIFF
--- a/bin/morse
+++ b/bin/morse
@@ -16,7 +16,7 @@ use strict;
 
 use Getopt::Std qw(getopts);
 
-my ($VERSION) = '1.3';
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: making case folding consistent

PR #950, opened 2025-02-12
morse: bad encoding

mknos:
* Fix incorrect encoding for symbol '(', found during testing
* Add encodings for symbols '@' and '_' (sync from NetBSD) [1]

1. https://github.com/NetBSD/src/blob/trunk/games/morse/morse.c

```
%perl morse -s '()' # previous identical encoding
-.--.-
-.--.-
```


briandfoy:
changes: Fix encoding (, add encodings for '@' and '_'

briandfoy (approved):

PR #951, opened 2025-02-19
moo: raise error for size>10

mknos:
* The maximum number of distinct digits is 10
* This version of moo takes an optional size argument
* It is more correct to show an error if the requested size is invalid
* Based on commit f3248c817c65b13363fbc27f5c10c11b7201475e in bin/fish, add entropy thanks to hash keys being "unordered"

### DIFF
--- a/bin/moo
+++ b/bin/moo
@@ -15,10 +15,20 @@ use strict;
 
 use List::Util qw(shuffle);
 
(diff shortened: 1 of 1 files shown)


briandfoy:
changes: raise error for size>10

briandfoy (approved):

PR #952, opened 2025-02-19
patch: improve validation of normal diffs

mknos:
* A normal diff can contain (a)dd, (c)hange and (d)elete commands
* For the Add command, a single target address is expected to be prefixed before "a"
* Diff utils output data in form of XaY,Z and not X,YaZ (so in practice the error condition should not happen)
* GNU diffutils manual described and Add command as "LaR", where L is a single line number and R is a range (e.g. 3,4) [1]
* Make this version of patch reject malformed Add commands
* Print the offending input data and line number
* Calling $patch->reject() is needed for setting the correct exit code
* When exiting on failure, avoid printing "done" message after preceding error messages
* I created an invalid input file manually to test this

1. https://www.gnu.org/software/diffutils/manual/html_node/Detailed-Normal.html

```
%cat YES.diff  
34,31a35
> new line
%perl patch -C -n yes YES.diff 
Checking patch against file yes using Plan C...
Malformed input at line 1: 34,31a35
1 out of 1 hunks ignored--saving rejects to .rej
%echo $?
1
```

### DIFF
--- a/bin/patch
+++ b/bin/patch
@@ -24,7 +24,7 @@ use constant EX_SUCCESS => 0;
 use constant EX_REJECTS => 1;
 use constant EX_FAILURE => 2;
 
-my $VERSION = '0.34';
+my $VERSION = '0.35';
 
 $|++;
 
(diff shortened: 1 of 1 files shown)


briandfoy:
changes: better error reporting for malformed add commands

PR #953, opened 2025-02-19
join: confusing declarations

mknos:
* Simplify code to make it clearer to the reader that get_arg() has only one function parameter, i.e. $arg
* Update loops using do-until-not pattern instead of regular do-while; found by perlcritic
```
%perl cat names # regression test
1 BoB
2 Budii
3 B-LITE
10 BLINK
%perl cat prices
1 $500
2 $30
3 50c
6 $2
%perl join names prices
1 BoB $500
2 Budii $30
3 B-LITE 50c
```

### DIFF
--- a/bin/join
+++ b/bin/join
@@ -132,12 +132,12 @@ while (@buf1 && @buf2) {
 if ($unpairables[0] && @buf1) {
   do{
     print_joined_lines(shift @buf1, [])
-  } until !get_a_line(\@buf1, $fh1);
+  } while get_a_line(\@buf1, $fh1);
 }
 if ($unpairables[1] && @buf2) {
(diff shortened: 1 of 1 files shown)


bin/join
```diff
@@ -132,12 +132,12 @@ while (@buf1 && @buf2) {
 if ($unpairables[0] && @buf1) {
   do{
     print_joined_lines(shift @buf1, [])
-  } until !get_a_line(\@buf1, $fh1);
+  } while get_a_line(\@buf1, $fh1);
```
briandfoy, line 135: wow, haven't seen `until` in awhile.

briandfoy (approved):

PR #958, opened 2025-02-24
grep -r should not sort directory entries

mknos:
* A directory might have 1000s of entries and sorting would make ```grep -r``` invocation slower (there is no standard option to toggle sort)
* Be consistent with OpenBSD and GNU versions by processing directory entries in the order they appear
* As part of verification, I compared output of ```grep -r``` on OpenBSD (filenames not sorted) with ```ls -1a``` (filenames sorted)
* Clarification: ```@list``` variable is only for setting up argument list for recursive matchfile() call. Declare it within if-block scope to avoid potential of old entries appearing between loop iterations.

```
%pwd
/home/x/PerlPowerTools
%# GNU filename matches are unsorted
%grep -r perl bin | cut -d ':' -f 1  | uniq | head -n 15
bin/uniq
bin/cal
bin/expand
bin/basename
bin/primes
bin/ls
bin/uname
bin/kill
bin/diff
bin/hexdump
bin/asa
bin/pwd
bin/whois
bin/expr
bin/patch
```

### DIFF
--- a/bin/grep
+++ b/bin/grep
@@ -58,7 +58,7 @@ use constant EX_MATCHED => 0;
 use constant EX_NOMATCH => 1;
 use constant EX_FAILURE => 2;
 
-our $VERSION = '1.018';
+our $VERSION = '1.019';
 
 $| = 1;                   # autoflush output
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: process directory entries in the order they appear instead of sorting them

PR #959, opened 2025-02-24
fortune: document the -v flag

mknos:
* Mention -v version number option in the usage string and POD manual
* Possibly the POD manual was copied from the NetBSD version, which doesn't have the -v option
* As done for other scripts, remove manual word breaks in POD which don't render nicely under pod2text
* Correct typos: "initial", "implementation"

### DIFF
--- a/bin/fortune
+++ b/bin/fortune
@@ -954,7 +954,7 @@ sub print_help
 
 	print <<EOF;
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: document the -v flag, and rewrap some docs

PR #971, opened 2025-02-26
cmp: 0X and 0x are valid hex prefix

mknos:
* Skip numbers (optional arguments 3 & 4) can be specified in decimal or hex
* When testing on OpenBSD and Linux, "0X" hex prefix is supported, so add it to this version too
* I didn't test on NetBSD, but the NetBSD version calls the C strotoll() function, which is documented to accept "0X" [1] [2]
* Tweak POD manual

1. https://cvsweb.netbsd.org/bsdweb.cgi/src/usr.bin/cmp/cmp.c?annotate=1.21  line128
2. https://man.netbsd.org/strtoll.3

```
%ifconfig > F1
%cp F1 F2
%perl cmp F1 F2
%perl cmp F1 F2 0x1 0X1 # same skip, different prefix, same day
%perl cmp F1 F2 0x1 0X2
F1 F2 differ: char 1, line 1
```

### DIFF
--- a/bin/cmp
+++ b/bin/cmp
@@ -219,7 +219,7 @@ exit $saw_difference;
 
 sub skipnum {
     my $n = shift;
-    return hex($n) if ($n =~ m/\A0x/);
+    return hex($n) if ($n =~ m/\A0[Xx]/);
     return int($n) if ($n =~ m/\A[0-9]+\z/);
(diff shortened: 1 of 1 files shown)


briandfoy:
changes: skip values can be decimal or hex; 0X and 0x are supported

briandfoy (approved):

Issue #965, opened 2025-02-25
bin/glob crash


mknos:
1. perl builtin glob() returns literal "{}" for input "{}"
```
%perl -MData::Dumper -e'warn Dumper(glob "{a,b}")'
$VAR1 = 'a';
$VAR2 = 'b';
%perl -MData::Dumper -e'warn Dumper(glob "{}")'
$VAR1 = '{}';
```
 
2. bash glob is consistent with perl builtin for this input
```
%echo $BASH_VERSION
5.2.15(1)-release
%echo {a,b}
a b
%echo {}
{}
```

3. standalone glob script fails for this input
```
%perl -Mwarnings -Mdiagnostics glob '{}'
Unescaped left brace in regex is illegal here in regex; marked by <-- HERE in
	m/\A{ <-- HERE }\Z/ at glob line 306 (#1)
    (F) The simple rule to remember, if you want to
    match a literal "{" character (U+007B LEFT CURLY BRACKET) in a
    regular expression pattern, is to escape each literal instance of it in
    some way.  Generally easiest is to precede it with a backslash, like
    "\{" or enclose it in square brackets ("[{]").  If the pattern
    delimiters are also braces, any matching right brace ("}") should
    also be escaped to avoid confusing the parser, for example,
    
     qr{abc\{def\}ghi}
    
    Forcing literal "{" characters to be escaped enables the Perl
    language to be extended in various ways in future releases.  To avoid
    needlessly breaking existing code, the restriction is not enforced in
    contexts where there are unlikely to ever be extensions that could
    conflict with the use there of "{" as a literal.  Those that are
    not potentially ambiguous do not warn; those that are do raise a
    non-deprecation warning.
    
    The contexts where no warnings or errors are raised are:
    
    as the first character in a pattern, or following "^" indicating to
    anchor the match to the beginning of a line.
    
    as the first character following a "|" indicating alternation.
    
    as the first character in a parenthesized grouping like
    
     /foo({bar)/
     /foo(?:{bar)/
    
    as the first character following a quantifier
    
     /\s*{/
    
Uncaught exception from user code:
	Unescaped left brace in regex is illegal here in regex; marked by <-- HERE in m/\A{ <-- HERE }\Z/ at glob line 306.
	FastGlob::recurseglob(".", "", "{}") called at glob line 272
	FastGlob::match_glob("{}") called at glob line 428
	FastGlob::glob("{}") called at glob line 449
```

[...]

PR #972, opened 2025-02-26
Reimplement glob

briandfoy:
Part of #965. This is a wrapper around File::Glob

### DIFF
--- a/bin/glob
+++ b/bin/glob
@@ -12,50 +12,140 @@ License: perl
 
 =cut
 
+package PerlPowerTools::glob;
+
+use strict;
+use warnings;
+
 our $VERSION = '2.1';
 
 use constant EX_SUCCESS => 0;
 use constant EX_FAILURE => 1;
+use constant EX_ERROR   => 2;
 
-use File::Glob qw(csh_glob);
+use File::Basename;
+use File::Glob qw(csh_glob GLOB_CSH);
 
-my $separator = "\n";
-if( $ARGV[0] eq '-0' ) {
-	shift @ARGV;
-	$separator = "\0";
+__PACKAGE__->run(@ARGV) unless caller();
+
+my %Unknown_users;
+my $wildcards = qr/[*?\[]/;
+
+sub run {
+	my( $class, @args ) = @_;
+
+	my( $code, $message ) = do {
+		if( @args == 0 ) {
+			( EX_ERROR, undef );
+			}
+		else {
+			my $separator = "\n";
+			if( $args[0] eq '-0' ) {
+				shift @args;
+				$separator = "\0";
+				}
+
+			my @ARGV_expanded = @args;
+			if( $^O =~ /\A MSWin32 \z/x ) {
+				@ARGV_expanded = map { expand_tilde($_) } @ARGV_expanded;
+				}
+
+			my @matches = csh_glob( "@ARGV_expanded", GLOB_CSH );
+
+			if( @matches ) {
+				$class->output_list( \@matches, $separator );
+				( EX_SUCCESS );
+				}
+			elsif( () = keys %Unknown_users ) {
+				( EX_FAILURE, $class->no_match_message );
+				}
+			else {
+				my $message = do {
+					my $pattern = "@args";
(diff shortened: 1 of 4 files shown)


bin/glob
```diff
@@ -5,45 +5,175 @@
...
+				}
+			elsif( () = keys %Unknown_users ) {
+				( EX_FAILURE, $class->no_match_message );
+				}
+			else {
+				my $message = do {
+					my $pattern = "@args";
+					my $unknown = () = keys %Unknown_users;
+
+					if( $^O ne 'MSWin32' && $pattern =~ /(?:\A|\s)~([\w-]+?)\b/ ) {
```
mknos, line 68: Variable $^O is tested in a slightly different way above (L50). One matches with spaces and the other doesn't.

mknos: Oops, I missed the /x suffix. But possibly having an is_windows() function is still useful.

bin/glob
```diff
@@ -5,45 +5,175 @@
...
 =head1 DESCRIPTION
 
-The B<glob> command/function implements globbing in perl, rather than
-forking a csh like Perl's built-in glob() call.  This is faster than
-the built-in glob() call, and more robust (on many platforms, csh
-chokes on C<echo *> if too many files are in the directory.)
+When this program was originally created, *perl* did not have a builtin
+C<glob> feature and would rely on the *csh* to do the work for it. With
+Perl v5.6 in March 2000, the L<File::Glob> module has done that work
+without interacting with *csh*.
```
mknos, line 169: This is a good change considering the age of the original script.

bin/glob
```diff
@@ -5,45 +5,175 @@
...
+	my $user = $1;
+
+	if( $user && ! $Unknown_users{$user}  ) {
+		my $net_user = `net user "$user" 2>&1`;
+		if( $net_user =~ /could not be found|The syntax of this command/ ) {
+			print STDERR "Unknown user $user.\n";
+			$Unknown_users{$user}++;
+			return;
+			}
+		s/ \A ~ (\w+) /$dir\\$1/x
```
mknos, line 111: I noticed the default value of $dir has slash, but this code to join $1 has a backslash. I'm not sure if that would make Windows confused.

mknos (approved):

mknos:
Quick test: "No match" message is printed without a newline

mknos:
Quick test2:
It's not considered an error if ~ cannot be expanded to a known username. Other glob versions fall back to treating "~" literally.

```
%for user in root notexist000; do echo "USER=$user"; echo -n "PERL -> "; perl -e'print glob("~" . shift . "/xx"),"\n"' "$user"; echo -n "BASH -> "; echo "echo ~$user/xx" > tmp.sh; bash tmp.sh; unlink tmp.sh; echo; done;
USER=root
PERL -> /root/xx
BASH -> /root/xx

USER=notexist000
PERL -> ~notexist000/xx
BASH -> ~notexist000/xx
```

briandfoy:
The previous script would fail if the tilde couldn't be expanded *and* there were special characters (`*` and so on) in the argument. With no special characters, the `~` is not expanded.

I'm following the csh glob function when I can, although the csh always tries to expand the tilde:

```
$ csh
% glob ~foo
Unknown user: foo.
% glob ~foo/*
Unknown user: foo.
```

briandfoy:
I think what I have now is good, but if you have additional test cases we can add them to *t/glob/glob.t*.

briandfoy:
@mknos Unless you have showstoppers, I'll merge this sometime this week.

mknos:
The main compat issue I found is for input '~nonuser00000/this/*'.
The old bin/glob matches perl builtin glob() and results in No Match.
The new code results in "Unknown user" error.
It's not really a showstopper; the current version could be merged.

briandfoy:
I aknowledge the behavior change, but I think the new behavior is correct for the original csh `glob` and that the PerlPowerTools version was wrong.

briandfoy:
changes: reimplement through File::Glob::csh_glob; this changes how ~ expansion works for unknown users, which will report an error if there is also a wildcard character.

PR #975, opened 2025-03-03
ls: runtime error for non-existent file argument

mknos:
* When testing ls today I discovered the error: Can't use an undefined value as a HASH reference at ls line 525.
* Test case: ```perl ls xxx ar a.x pig```  ...where files xxx and a.x don't exist
* I broke this in the previous commit around where DirEntries() is called for ```@Files``` list
* Add the correct return value guard
* DirEntries() prints a warning and returns an empty list if stat() fails on its file argument
* Avoid listing filenames where stat() failed by storing good arguments in ```@okfiles``` list

```
%perl ls  xxx ar a.x pig # with change applied
ls: can't access 'xxx': No such file or directory
ls: can't access 'a.x': No such file or directory
ar   pig
```

### DIFF
--- a/bin/ls
+++ b/bin/ls
@@ -522,9 +522,10 @@ if ($#ARGV < 0) {
 		my %attrs;
 		foreach (@Files) {
 			my @ret = DirEntries(\%Options, $_);
+			next unless @ret; # stat() failed
 			%attrs = (%attrs, %{ $ret[-1] });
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: avoid listing filenames where stat() fails

PR #976, opened 2025-03-03
grep: documentation update

mknos:
* In pod SYNOPSIS, -i was listed twice and -Z wasn't listed
* Usage string was missing -Z too
* Sort single letter options, including getopts() call; this makes it easier to see what is happening
* No intended functional change

### DIFF
--- a/bin/grep
+++ b/bin/grep
@@ -92,35 +92,35 @@ sub VERSION_MESSAGE {
 
 sub usage {
 	die <<EOF;
-usage: $Me [-IincwsxvHhLlFguRraqT] [-e pattern] [-A NUM] [-B NUM] [-C NUM]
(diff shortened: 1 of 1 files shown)


bin/grep
```diff
@@ -92,35 +92,35 @@ sub VERSION_MESSAGE {
...
+	-C   show lines of context around each matching line
 	-c   give count of lines matching
-	-w   word boundaries only
-	-q   quiet; nothing is written to standard output
-	-x   exact matches only
-	-v   invert search sense (lines that DON'T match)
-	-H   show filenames
-	-h   hide filenames
 	-e   expression (for exprs beginning with -)
+	-F   search for fixed strings (disable regular expressions)
```
briandfoy, line 105: thank you for alphabetizing this! I hate when these sorts of docs don't have an apparent sequence.

briandfoy (approved):

PR #977, opened 2025-03-04
chgrp: fail early for empty string

mknos:
* Similar to commit 471e64b3e0ec3a1fd9c85ecec085d97fabe47417 for bin/chown
* Fail early if the 1st (group) argument is an empty string to avoid unnecessary getgrnam('') call
* style: forward declaration of modify_file() is not needed if the method is called with parens
* style: move default exit() statement above the subroutine declarations, as done in other scripts

### DIFF
--- a/bin/chgrp
+++ b/bin/chgrp
@@ -11,8 +11,6 @@ License: perl
 
 =cut
 
-
-
 use strict;
 
 use File::Basename qw(basename);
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: fail if the group is the empty string

PR #978, opened 2025-03-04
tar: fail early for bad options

mknos:
* Add missing getopts() return value check
* This version of tar didn't have a usage string, so add one
* Don't list -c in usage string for now because archive creation isn't implemented
* Unconditionally use Compress::Zlib because it is a core library

```
%perl tar -x -f yo -L
Unknown option: L
usage: tar {-tx} [-mvZz] [-f archive] [file ...]
```

### DIFF
--- a/bin/tar
+++ b/bin/tar
@@ -13,6 +13,7 @@ License:
 
 use strict;
 
+use Compress::Zlib qw(gzopen);
 use File::Basename qw(basename);
 use Getopt::Std;
 use IO::File;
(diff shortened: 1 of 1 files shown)


bin/tar
```diff
@@ -13,34 +13,21 @@ License:
 
 use strict;
 
+use Compress::Zlib qw(gzopen);
```
briandfoy, line 16: This is a v5.10 module, so we can't rely on it being there. The conditional check  should still be there,.

briandfoy (changes requested):

bin/tar
mknos: Thanks for the feedback. New commit added which restores the module-exists check, without conditionally changing the personality of getopts().
As a test I created an archive with GNU tar 1.34, listed the entries with ```perl tar -tf x.tar```, compressed the archive with gzip 1.12 and did a detailed list of compressed archive with ```perl tar -tvzf  x.tar.gz```.

bin/tar
```diff
@@ -225,7 +211,8 @@ else
 
   if ($opt{'z'} || $opt{'Z'})
    {
-    # quick and dirty till we sort out Compress::Zlib
+    eval { require Compress::Zlib } or fatal('Compress::Zlib not found');
```
briandfoy, line 214: yeah, good change.

briandfoy (approved):

PR #980, opened 2025-03-05
install: -m MODE validation

mknos:
* Passing an empty string as a mode was being accepted and interpreted as -m0
* Empty string is not a valid mode specifier so catch it early
* I found this when testing against GNU and OpenBSD versions


briandfoy:
changes: catch empty argument to -m

briandfoy (approved):

PR #981, opened 2025-03-05
tac: don't print version number in usage string

mknos:
* If I type a bad option I want to see the correct options, not the program version
* As done in other scripts using getopts(), support --version option by adding a VERSION_MESSAGE() function


briandfoy:
change: don't print version number in usage string

briandfoy (approved):

PR #982, opened 2025-03-05
chmod: show which option is wrong

mknos:
* The supported options are -R/-L/-H/-P, and grouping of options is allowed (e.g. -LP)
* When adding an unexpected option letter to a group, the error message was misleading
*  ```perl chmod -Rp``` reported "invalid option -- pR", but R is a valid option
* Fix this by validation the option letters after the option string has been split into a single letters

```
%perl chmod -Rp 0 0 0 0 0 0 0 0 0 0
chmod: invalid option -- 'p'
usage: chmod [-R [-H | -L | -P]] mode file...
```

### DIFF
--- a/bin/chmod
+++ b/bin/chmod
@@ -30,12 +30,12 @@ while (@ARGV && $ARGV [0] =~ /^-/) {
     my $opt = reverse shift;
     chop $opt;
     last if ($opt eq '-');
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: show which options were invalid

PR #984, opened 2025-03-06
install: -o0 and -g0 ignored

mknos:
* Include options -o and -g in the empty-string validation previously added for option -m
* Generally uid 0 means "root user" on unix, and -o 0 could be given to explicitly install files as this user
* The call to chown() was bypassed because of truth check
```
%perl mkdir d0 # after patch, as non-root user
%perl install -o0 file d0
install: chown 0.1000 d0/file: Operation not permitted
%perl install -g0 ar d0
install: chown 1000.0 d0/ar: Operation not permitted
```

### DIFF
--- a/bin/install
+++ b/bin/install
@@ -54,9 +54,11 @@ if ($opt{d} and grep($_, @opt{qw/ C c D p /}) > 0) {
     warn "$Program: -d not allowed with -[CcDp]\n";
     usage();
 }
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy (approved):

briandfoy:
changes: proper validation for -g and -o

Issue #988, opened 2025-03-14
ls does not print 'l' mode indicator for symbolic links


mknos:
* In long-listing mode (-l), the ls command prints the file mode as the first element on each line
* I created a symlink AP which refers to the file "apply"
* The GNU version prints "l" prefix on the file mode string but the perl version prints '-'
* The standards document says the 1st character of the mode-string is the entry-type, and for a symbolic link the entry-type should be "l" ("-" should be printed only for regular files) [1]
* I also verified that the OpenBSD and Sortix versions of ls print the "l" entry-type

1. https://pubs.opengroup.org/onlinepubs/9699919799/utilities/ls.html

```
%readlink AP
apply
%ls -l AP # GNU
lrwxrwxrwx 1 x x 5 Mar 14 10:42 AP -> apply
%perl ls -l AP
-rwxr-xr-x   1 x       x            4488 Jan 19 14:25 AP
```

mknos:
The file size printed for AP indicates the symlink might be followed before the entry listing occurs. The -l option is supposed to list symlinks as is.

briandfoy:
I have an initial fix in #989, but I need to check that I didn't break something else. Feel free to build on that if you want to play with it.

briandfoy:
We're going to need better tests for *bin/ls* because the fix I tried probably breaks other things. 

In every case, *bin/ls* calls `stat` on its filename argument, so it always gets the target. If I change that to something like this is works out:

    $mode = -l $Arg ? lstat($Arg) : stat($Arg);

However, there are several places where I make this change and I don't know if they are all appropriate; they probably aren't.

Then, in `format_mode`, the mode string is constructed just from the number, with the highest three bits used for the file type as an index into the `@ftype` array. For a link, that should be `0b101`, but a regular file is `0b100`. 

But, there's are several other problems where the *bin/ls* does not check for a symlink, so it does the wrong thing for dangling symlinks (the `@` is a macOS thing). This dangles because *000.link* points to *t/000.basic.t*, so that would resolve to *t/t/000.basic.t*:

```
$ ls -l t/000.link
lrwxrwxr-x@ 1 brian  staff  13 Mar 14 19:11 t/000.link -> t/000.basic.t
$ perl bin/ls -ld t/000.link
ls: can't access 't/000.link': No such file or directory
```

*bin/ls* should not care if this resolves.

[...]

PR #989, opened 2025-03-14
Respect symlinks in bin/ls

briandfoy:
See #988

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -154,8 +154,10 @@ t/echo/echo.t
 t/factor/factor.t
 t/false/false.t
 t/find/find.t
+t/glob/glob.t
 t/lib/common.pl
 t/lib/utils.pm
+t/ls/ls.t
 t/mimedecode/000.basic.t
 t/nl/nl.t
 t/od/od.t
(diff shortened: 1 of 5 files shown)


mknos:
Thanks for taking a look at this.
I fetched a copy of bin/ls and ran it within the repo's bin/ directory.
When I run ls with no arguments, the number of columns in the output is different compared to the previous version.
In this teset my xterm has width of 94, the old ls prints 6 columns of output but this version only prints 5. It looks like there is space for the 6th column but no text is filled into it.

briandfoy:
Yes, the number of columns is reduced by one because the math happened to not work on the one directory I tried with the particular terminal width I had. I think it's an off by one error somewhere. As such, I took back one column simply to keep it from wrapping the newline.

mknos:
OK. I recall the the `$Maxlen+=1` statement in List() which is supposed to account for a space before the next column starts. But what happens about the last column? It might not be correct to increment $Maxlen for the last column because only newline will be printed after the filename, not another column.

mknos:
The sprintf() format string has a trailing space which is not part of $Maxline; however, $Maxline has already accounted for one space. That space makes it more difficult to see what's happening.

PR #990, opened 2025-05-02
cal: correct current month value

mknos:
* When running cal with no arguments the month number is determined from localtime(), which returns 0 for Jan
* Function fmt_month() needs 1 for Jan, so adjust the value passed in
* Patch tested against /usr/bin/cal on ubuntu (from package bsdmainutils)
* test1: ```perl cal 0 2025``` ---> invalid month
* test2: ```perl cal``` ---> now shows current month (5)
* test3: ```perl cal 5 2025```  ---> same output as test2


briandfoy:
changes: fix off by one error in month

briandfoy (commented):
good catch

PR #992, opened 2025-11-01
Add test for 'yes'

ludovic-tc:
This PR adds a test for the `/bin/yes` script.

I hope this may be considered useful even though 'yes' is a very simple program.  My intention was partly to complete an easy task with a view to hopefully making more substantial contributions in future.

I have attempted to write it cogently and adapt it to the house style, but suspect I have come up short on both counts.

I could not find a standing dev branch so have targeted master, but can easily target a different branch if desired.

Should this humble contribution is of interest to the project, I would be honoured to receive any and all feedback, whether detailed or e.g. just 'go back and look at these guidelines again' or 'run perltidy with this config'. 

In particular, I will quite understand if you would prefer this rebased into a single commit. 

I have run it with `prove` from my forked repo's root, and it seemed to play well with the other tests, but do let me know if I have missed some complete howlers. 

Best wishes and thanks.


ludovic-tc:
Hmm – looks like the Appveyor ubuntu2004 build failed! That's a surprise ... I'm guessing the child process got blocked somehow. I specifically wrote it to perform a unidirectional read so I didn't expect that to be a problem.

briandfoy (changes requested):
The test looks good, but I think you have an extra file in this commit. Did you mean to include *t/yes/yes*?

briandfoy (approved):

briandfoy:
This will be part of the next release, which might be a bit. Thanks,

PR #995, opened 2025-12-02
mkdir: Suppress extra newline in warning messages

mknos:
Strings passed to die() and warn() already had an explicit newline, so there's no need for the SIG handlers to add it.

```
$ perl mkdir -m 0x0 newdir # example die()
mkdir: invalid mode: 0x0
$ perl mkdir -m '' newdir # example warn()
mkdir: invalid mode
$
```


briandfoy:
changes: handle line endings in error messages

PR #999, opened 2025-12-03
uniq: avoid double negative in POD

mknos:
* The existing text is confusing
* Follow the GNU manual and use "duplicate lines" because "duplicate" is how I remember what `uniq -d` does


briandfoy (approved):
much better

PR #1000, opened 2025-12-03
head: correct error message for -n

mknos:
* If the argument to -n is an empty string, display the "invalid number" error
* Previously comparison with zero resulted in a "too small" error
* Found when testing against GNU version


briandfoy:
There it is: issue #1000 :)

briandfoy:
changes: correct error message for -n

PR #1008, opened 2026-01-10
Fix addbib to handle ^D for Abstract field

briandfoy:
The addbib program has to be in raw mode so it can handle ^D on its own for the abstract field. This ended up completely replacing the existing program.

This requires Term::ReadKey, which is not a core module. This violates the rules of PerlPowerTools, but this  program has been in the list for 20 years, so I don't want to take it out.

See #1007

Co-Authored-By: zongnan.hu@gmail.com

### DIFF
--- a/t/addbib/addbib.t
+++ b/t/addbib/addbib.t
@@ -4,8 +4,20 @@ use strict;
 use Test::More 0.95;
 
 BEGIN {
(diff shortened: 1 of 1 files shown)


briandfoy:
changes: handle ^D properly

PR #1009, opened 2026-01-12
Add tests and examples for apply

samuel-k-hu:
Hi Brian,

I’ve been exploring the PerlPowerTools codebase as part of learning Perl, and while doing so I noticed that apply doesn’t currently have dedicated tests or examples.

This PR adds a set of tests for apply, as well as documentation examples. Specifically, the tests make use of apply’s debug mode to compare the actual commands that apply would execute with the commands we expect it to execute. Based on this comparison, the tests check whether apply behaves as intended.

I’m still relatively new to Perl, so I would very much appreciate any feedback or suggestions on how these tests or the documentation could be improved or made more idiomatic.

### DIFF
--- a/bin/apply
+++ b/bin/apply
@@ -171,6 +171,28 @@ of the arguments is handed to a shell for interpretation of meta characters.
 This shell is determined by Perl, during its building phase, and is typically
 C</bin/sh -c> on Unix platforms.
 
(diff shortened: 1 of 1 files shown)


briandfoy:
Very nice, thanks,

PR #1010, opened 2026-01-12
Fix addbib tests on Windows and Cygwin

samuel-k-hu:
Hi Brian,

First of all, thank you for maintaining `addbib`. After some testing, I found that addbib itself works well, but not all tests pass reliably across platforms. I believe the issues are with the tests rather than with the program logic.

On Windows, the tests related to “saw instructions” fail, even though in my experiments the instructions are actually displayed correctly. In addition, testing unwriteable files is not reliable on Windows. For this reason, I think it is reasonable to skip these tests on Windows for now.

On Cygwin, the unwriteable-file test also fails, but for a different reason. Although /etc exists under Cygwin, files there are generally writeable by normal users, which causes the test to behave differently from what is expected on typical Unix systems. To make the test more robust, I changed it to create an explicitly unwriteable file specifically for testing.

With these changes, the test suite behaves more consistently across platforms and avoids false failures on Windows and Cygwin.

### DIFF
--- a/t/addbib/addbib.t
+++ b/t/addbib/addbib.t
@@ -21,10 +21,20 @@ BEGIN {
 		done_testing();
 		exit;
 		}
+
+	if ( $^O eq 'MSWin32' ) {
+		system 'attrib +r ./t/data/unwriteable_file.txt';
+		}
 	}
 
 BEGIN {
 	*CORE::GLOBAL::exit = sub { $_[0] // 0 }
+	}
+
+END {
+	if ( $^O eq 'MSWin32' ) {
+		system 'attrib -r ./t/data/unwriteable_file.txt';
+		}
 	}
 
 use lib qw(t/lib);


t/addbib/addbib.t
```diff
@@ -8,8 +8,8 @@ BEGIN {
 		if( ! eval "require Term::ReadKey" ) {
 			'Term::ReadKey required for testing'
 			}
-		elsif( $^O eq 'MSWin32' and $ENV{'GITHUB_ACTIONS'} ) {
-			q(Can't test on Windows in GitHub Actions);
+		elsif( $^O eq 'MSWin32' ) {
```
briandfoy: Instead of blocking all of MSWin32, I'd like to see what we can fix. I'd put this back becasue it has no hope of working under CI since there is no terminal.

We can add additional checks for Windows and Cygwin.

But, I'd really like to get this working on Windows.

samuel-k-hu:
Exactly. The negative tests are actually failing, but the failure happens so early that it produces no output—which coincidentally mimics the expected successful result.

I've tested this in both `cmd.exe` and `PowerShell`, and the behavior is consistent across both.

I feel like I need to catch up on quite a bit of foundational knowledge to properly solve this. I've been trying to run the script without using `run_command`, but I haven't been successful yet.

briandfoy:
Yeah, I'm really weak on Windows, but I'll look at this again.

I had the same problem trying to test *bin/bc*. 

So, leave this pull request as it is, and let me think about it for a bit (or if you figure it out, that's fine too). Maybe it is just untestable on Windows.

PR #1014, opened 2026-01-13
ar: add test files

samuel-k-hu:
I have created several files for testing and set up a basic code framework.

However, the current tests are failing because the files produced by PerlPowerTools ar differ from those generated by GNU `ar` and FreeBSD `ar`.

I believe the overall logic of the code is correct, but there are some differences in how certain details are handled, particularly with respect to the archive format and the treatment of original file metadata.

Below are the results produced by the same command(`ar q out01.a a b c`) using three different implementations of ar:

**PerlPowerTools**

```
!<arch>
a               1768312629  197609197121100660  2         `
a
b               1768312629  197609197121100660  2         `
b
c               1768312629  197609197121100660  2         `
c
```

**FreeBSD**

```
!<arch>
a/              0           0     0     100644  2         `
a
b/              0           0     0     100644  2         `
b
c/              0           0     0     100644  2         `
c
```

**GNU ar**

```
!<arch>
a/              1768312629  197609197121100660  2         `
a
b/              1768312629  197609197121100660  2         `
b
c/              1768312629  197609197121100660  2         `
c
```

The `out*.a` files currently in the repository were generated using GNU `ar`.

I understand that this commit will require further changes, but I am submitting it now to share the current progress of this work.

### DIFF
--- a/t/ar/ar.t
+++ b/t/ar/ar.t
@@ -28,7 +28,7 @@ subtest 'ar q out01.a a b c' => sub {
 	my ( $fh, $filename ) = File::Temp::tempfile();
 	my $result = run_command( $program, [ 'q', $filename, $a_f, $b_f, $c_f ], undef );
 	is $result->{'exit'},              0, 'exited successfully';
-	is compare( $filename, $out01_f ), 0, 'execution succeeded';
+	TODO: { local $TODO = 'not working yet'; is compare( $filename, $out01_f ), 0, 'execution succeeded'; }
 	};
 
 


samuel-k-hu:
I did some literature review. The `ar` utility has at least two styles of formats ([BSD and SVR4/GNU](https://man.freebsd.org/cgi/man.cgi?query=ar&apropos=0&sektion=5&manpath=FreeBSD+15.0-RELEASE&arch=default&format=html)), but the standard does not specify which format should be used by default. Based on testing, FreeBSD’s ar seems to use the SVR4/GNU format. OpenBSD appears to have [adopted GNU ar around version 4.0](https://man.openbsd.org/OpenBSD-4.0/ar.1) and later [replaced it with llvm-ar](https://man.openbsd.org/OpenBSD-7.8/ar.1). [FreeBSD 15 also seems to use llvm-ar](https://man.freebsd.org/cgi/man.cgi?query=ar&apropos=0&sektion=1&manpath=FreeBSD+15.0-RELEASE&arch=default&format=html). By default, llvm-ar [writes in the SVR4 format but can read BSD-format archives](https://releases.llvm.org/3.5.0/docs/CommandGuide/llvm-ar.html). This suggests that, currently, most implementations of ar default to reading and writing in the SVR4/GNU format, while still being able to read BSD-format archives. 

According to test results, the `ar` in PerlPowerTools only supports the BSD format. This means that other versions of `ar` can handle files packaged by PerlPowerTools, but PerlPowerTools may not correctly handle files packaged by other ar implementations.

I think it is at least necessary for `ar` to support both formats. Perhaps it could follow the approach of llvm-ar, using one format as the primary read/write format while still being able to read the other format.

briandfoy (approved):
I left some notes, but this is a good start.

briandfoy:
I fixed a problem in the addbib test and merged your changes, so you should rebase on master.

briandfoy:
I made the one failing test `TODO` so it doesn't stop the tests.

briandfoy:
changes: add tests

PR #1017, opened 2026-01-16
ar: add support for reading SVR4 variant files

samuel-k-hu:
Hi Brian,

Previously, `ar` could not correctly list file names in SVR4 variant archives. I renamed the original readAr to readArBSD and implemented `readArSVR4`, then made `readAr` dispatch to the appropriate function based on the archive type. There is some duplication between `readArSVR4` and `readArBSD`. In the future, I plan to extract shared logic into `readAr`, but for now I prefer to leave them as-is since I’m still becoming familiar with the `ar` code. This approach allows me to minimize changes to the existing source.

During this process, I also noticed that ar cannot correctly handle multiple files with the same name but different contents, which differs from FreeBSD’s ar.The two also handle file metadata very differently. Achieving full compatibility between PerlPowerTools’ ar and FreeBSD’s version will likely require substantial further work.

Any suggestions are welcome.

### DIFF
--- a/bin/ar
+++ b/bin/ar
@@ -13,7 +13,6 @@ License: perl
 
 use strict;
 
-use bytes;
 use POSIX qw(strftime);
 use File::Basename;
 use FileHandle;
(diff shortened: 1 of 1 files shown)


briandfoy (changes requested):
1. Update MANIFEST
2. I have a question about your use of `bytes`

PR #1018, opened 2026-01-18
which: add silent flag

mknos:
* Silent mode (-s) is supported on Linux, FreeBSD, DragonflyBSD and Mac
* The exit code for -s is identical but nothing is printed
* This implementation follows OpenBSD by printing a warning for each missing command; suppress it for -s mode
* Bump version
* Stop polluting the usage string with the program number; add --version


briandfoy:
changes: add flag for silent mode

PR #1032, opened 2026-03-07
enable wildcards in ls and rm

samuel-k-hu:
Hi Brian, 

I recently discovered that [rename](https://metacpan.org/dist/File-Rename) works with wildcards on Windows. So I looked into its code and found it uses a very simple method. I've integrated it into `ls` and `rm`, and it seems to perfectly meet my needs.

Have a nice day!

### DIFF
--- a/bin/ls
+++ b/bin/ls
@@ -529,6 +529,7 @@ sub run {
 	my $Options;
 	( $Options, @args ) = $class->process_options(@args);
 	@args = qw(.) unless @args; # current directory if nothing else
+	@args = map {glob} @args if $^O =~ m{Win}msx;
 
 	my $ArgCount = -1;  # file/directory argument count
 	my %Attributes;     # File::stat directory entry attributes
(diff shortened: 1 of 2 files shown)


briandfoy (changes requested):
See the comments in the pull request.

samuel-k-hu:
Hi Brian,

Thank you for your comment. As I understand it, does this mean that I can introduce a common module into PerlPowerTools and have all programs use the subroutines in that module to handle `@ARGV`?

briandfoy:
As I've been working on this in #1025, I've been trying all the weird situations and found another wrinkle I hadn't really thought about. If the glob does not return any filenames, it uses the pattern as the literal result. So, we can't just call `glob` and use its result. In the zero-result case, we have to return the pattern. 

Since I'm working on it in the other PR, and it's actually a bit of work to do what BSD glob actually does, I'm going to close this one.

PR #1034, opened 2026-06-03
nl: show usage for invalid -b0

mknos:
* Truth check was causing nl -b0 to be ignored, but 0 is not an expected value
* Follow the pattern that already existed for option -s, also for other options setting a default value like this


briandfoy (approved):

briandfoy:
changes: properly set default values for options

PR #1035, opened 2026-06-03
what: print filenames unconditionally

mknos:
* Print name once per file argument, including files containing zero matches
* Empty filename '' is generally invalid and would fail earlier in open()

```
%perl what /bin/gcc /bin/gprof ''
/bin/gcc:
/bin/gprof:
	 Copyright (c) 1983 Regents of the University of California.
Unable to open '': No such file or directory
```

### DIFF
--- a/bin/what
+++ b/bin/what
@@ -44,7 +44,7 @@ sub printWhat
     my $file_name   = shift;
     my $stop_flag   = shift;
     my $line;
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: don't need to test the filename for length

PR #1037, opened 2026-06-04
glob: terminate error strings with newline

mknos:
* Message strings passed into $class->exit() did not include a newline---print one here instead of updating all the strings
* This was found when testing on Linux

### DIFF
--- a/bin/glob
+++ b/bin/glob
@@ -90,7 +90,7 @@ sub run {
 
 sub exit {
 	my( $class, $code, $message ) = @_;
-
+	chomp $message;
 	print { *STDERR } "$message\n" if defined $message;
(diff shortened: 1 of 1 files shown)


briandfoy (changes requested):
When I do this (usually, heh), I first chomp the message and then re-add the newline. If the message has no newline, `chomp` does nothing.  If there is a newline, it's removed and added back.

I've been very bad about deciding who should put the newline there and have two projects that do it in opposite ways.

briandfoy:
@mknos - if you are fine with my addition, I'll merge this.

briandfoy:
changes: ensure all messages end in a newline

PR #1038, opened 2026-06-05
tac: remove user-configurable buffer size option

mknos:
The "-S size_bytes" option does not appear in other versions of tac and it seems over-complicated to include one here. The NetBSD and DragonFlyBSD versions of tac do not support any options. Probably the reference implementation for initial development was GNU, but that only has -b, - r and -s sep_value. Technically the -B flag isn't provided by GNU tac either, but I decided to leave it.

Default buffer size becomes a constant so it is easy to customise.

Regression test...
```
%/usr/bin/tac --version
tac (GNU coreutils) 8.30
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Jay Lepreau and David MacKenzie.
%/usr/bin/tac tac  > a && perl tac tac > b && perl cmp a b 
%
```

### DIFF
--- a/bin/tac
+++ b/bin/tac
@@ -11,11 +11,6 @@ License: perl
 
 =cut
 
-
-#
-# tac - concatenate and print files in reverse
-#
-
 use strict;
 
 use File::Basename qw(basename);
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: remove non-standard -S switch

Issue #1036, opened 2026-06-03
apply enters infinite loop for %0 pattern


mknos:
Hello,

When testing apply(1) tonight I discovered the following test:
```
%perl apply -d 'ls -a %0' /dev
exec ls -a /dev
exec ls -a /dev
exec ls -a /dev
exec ls -a /dev
exec ls -a /dev
exec ls -a /dev
exec ls -a /dev
exec ls -a /dev
exec ls -a /dev
exec ls -a /dev
exec ls -a /dev
exec ls -a /dev
exec ls -a /dev
exec ls -a /dev
exec ls -a /dev
exec ls -a /dev
exec ls -a /dev
exec ls -a /dev
exec ls -a /dev
exec ls -a /dev
exec ls -a /dev
exec ls -a /dev^C
```

There are two problems with this. Firstly, %0 is not a valid substitution as far as apply(1) can understand. The BSD version does not attempt to substitute it, but the Perl version incorrectly does. Secondly, the Perl version does not remove /dev from the  remaining arguments list after it is used.

When testing on NetBSD, the output of the above is...
```
exec ls -a %0 /dev
```

As a secondary test on NetBSD, I add an extra directory argument.
```
%apply -d 'ls -a %0' /dev /etc
exec ls -a %0 /dev
exec ls -a %0 /etc
```
From this, the expected behaviour is as if the -1 option was given, i.e. construct and run one new command per argument.
I don't have a fix at this time.

mknos:
Something else I forgot to note. The NetBSD manual states that %NUMBER substitutions allow for only a single, non-zero digit: [1]
```
On each execution of command, each character sequence of the form ``%d''
     in command, where d is a digit from 1 to 9
```
The Perl version allows multiple digits to be given, e.g. `%12`. I don't think it's necessary to restrict this, but validation logic would be needed for
1) not substituting `%0`
2) possibly not substituting, e.g. `%7` if there weren't enough arguments.

The current ARGV array index calculation might end up being wrong because of negative values wrapping around to the end of array.

1. https://man.netbsd.org/apply.1

briandfoy:
I say we not substitute `%0`, and for the case of not enough arguments, I don't have a suggestion.

briandfoy:
Looking at this, I just need to change the regex that checks for `%\d` in the command string to decide how many things to grab off the argument list. When it's `%0` only, it takes nothing off the argument list, so the argument list never decreases, hence the infinite loop.

[...]

PR #1039, opened 2026-06-05
allow only 1 to 9 for magic (#1036)

briandfoy:
Be more strict about magic in apply (#1036)


briandfoy:
changes: apply only allows 1-9 for magic in command strings. 0 creates an infinite loop.

PR #1041, opened 2026-06-07
tail: fix bug in rewriting argv

mknos:
Function new_argv() handles translation of bare [-+]x option. This failed to identify if the pattern appeared immediately after "-n". This was found when testing against the GNU version.

Tests:
* perl tail -n 2 tail # last 2 lines
* perl tail -n -2 tail # last 2 lines
* perl tail -n +2 tail # show line 2 and all subsequent lines

### DIFF
--- a/bin/tail
+++ b/bin/tail
@@ -74,6 +74,7 @@ sub new_argv
 {
     my @new;
     my $end = 0;
+    my $prev = '';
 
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
changes: recognize `-\d` after a `-n`

PR #1044, opened 2026-06-11
head: allow '+' prefix on line count

mknos:
* The standards document says `head -n COUNT` counts from one [1]
* The OpenBSD and GNU versions accept a '+' prefix on a number, so allow that here for compat

Tests:
perl head -2 head --> 2 lines
perl head -n2 head --> 2 lines
perl head -n+2 head --> 2 lines
perl head -n-2 head --> invalid

1. https://pubs.opengroup.org/onlinepubs/9799919799/utilities/head.html

### DIFF
--- a/bin/head
+++ b/bin/head
@@ -21,7 +21,7 @@ use constant EX_SUCCESS => 0;
 use constant EX_FAILURE => 1;
 
 my $Program = basename($0);
-my ($VERSION) = '1.3';
+my ($VERSION) = '1.4';
 
 @ARGV = new_argv();
 my %opt;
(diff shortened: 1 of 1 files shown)


bin/head
```diff
@@ -32,11 +32,11 @@ unless (getopts('n:', \%opt)) {
 my $count;
 if (defined $opt{'n'}) {
     $count = $opt{'n'};
-    if (length($count) == 0 || $count =~ m/[^0-9]/) {
+    if ($count !~ m/\A[\-\+]?[0-9]+\Z/) {
         warn "$Program: invalid number '$count'\n";
         exit EX_FAILURE;
     }
-    if ($count == 0) {
+    if ($count <= 0) {
```
briandfoy, line 39: Do we accept the `-n-2` syntax because it's easier to handle the invalid syntax here with the `<=`?

briandfoy (changes requested):

bin/head
mknos: I think the main reason to do this was to show an out-of-range error for negative numbers, instead of the previous not-a-number error. As an extension to the standard, the GNU version allows a negative number for -n which chops COUNT lines from the end of input. I've never used that feature on Linux, but if someone used it here they'd see out-of-range. This might encourage the user to look at the POD instruction about it needing to be positive.

briandfoy: Fair enough. Just wanted to check.

briandfoy:
changes: count from 1 and accept +N syntax

PR #1045, opened 2026-08-20
apply: forbid multiple -# argument

mknos:
Following the OpenBSD version, raise an error for bad usage of conflicting (or multiple of the same)  -# option. Tested on Linux and Windows.

* Test case (invalid): `perl apply  -2 -3 echo 1 2 3 4 5 6`
* Previously the final -# option would win (-3); now print an error and exit


briandfoy (approved):
A bit klunky but it gets the job done without rewriting everything

PR #1047, opened 2026-08-20
ar: remove manual hyphenation in POD

mknos:
The perldoc and pod2text commands perform their own formatting of POD text. On my system the manually hyphenated text did not render correctly because the two parts of the word appeared on the same line (hell- o). Fix this manually, checking that column width fits inside 80 characters. No wording change.

This was found with...
```
pod2text ar | aspell -a | sort | uniq | less
```


briandfoy (approved):

briandfoy:
changes: don't hyphenate words for line breaks since the pod reader can do that

PR #1049, opened 2026-08-20
ed: helper function to flag buffer as modified

mknos:
Sub-commands in ed which potentially modify the buffer will mark it as dirty so the "q" sub-command can warn about unsaved changes. While here, I decided to express maxline() in terms of List::Util::max(). No intended functional change.


briandfoy (approved):

briandfoy:
changes: refactor file changed flags for consistency

PR #1050, opened 2026-08-27
patch: fix typo "defined"

mknos:
I found error today when running patch program on Ubuntu.
I checked that no other files had that same typo.
```
% perl -Mdiagnostics patch ar < ar.diff
Can't call method "defiend" on unblessed reference at patch line 378, <STDIN>
        line 5 (#1)
    (F) A method call must know in what package it's supposed to run.  It
    ordinarily finds this out from the object reference you supply, but you
    didn't supply an object reference in this case.  A reference isn't an
    object reference until it has been blessed.  See perlobj.

        Can't call method "defiend" on unblessed reference at patch line 378, <STDIN> line 5.
        Patch::bless(Patch=HASH(0x614948aec970)) called at patch line 133
```

### DIFF
--- a/bin/patch
+++ b/bin/patch
@@ -375,7 +375,7 @@ sub bless {
     my $class = "Patch::\u$type";
 
     my ($options, $garbage) = @{$_[0]}{'options', 'garbage'};
-    $options = defiend $options ? $options : [];
+    $options = defined $options ? $options : [];
 
     # New hunk, same patch.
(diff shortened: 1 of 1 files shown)


briandfoy (approved):

briandfoy:
okay, this is scary. How did I not catch this in the tests?

briandfoy:
changes: fix syntax error from mistyped keyword

PR #1054, opened 2026-09-03
cut: use error exit-code if a file argument does not open

mknos:
* When testing cut today I discovered that it uses the idiomatic while(<>) loop for reading input in handle_b() and handle_f()
* If file arguments are provided, all should process correctly in order for a success exit-code
* Failure to open a file argument means process-next-file, not terminate-program
* Hijack the WARN signal to set exit code; in future handle_b() and handle_f() could be restructured if necessary
* If a directory is passed as an argument it is ignored; this is consistent with OpenIndiana & OpenBSD, but GNU cut raises a warning for it
* For now, no attempt is made to suppress line numbers in the warning message

Test from "Git-Bash"...
```
$ perl -v | perl grep -F built
This is perl 5, version 42, subversion 2 (v5.42.2) built for x86_64-cygwin-thread-multi
$ echo 'y,a,y' > cheer
$ perl cut.pl -c1,2  nothing .. cheer
cut.pl: Can't open nothing: No such file or directory at cut.pl line 122.
y,
$ echo $?
1
```

### DIFF
--- a/bin/cut
+++ b/bin/cut
@@ -47,7 +47,7 @@ elsif (defined $opt{'f'}) {
     handle_f($opt{'f'}, $opt{'d'}, $opt{'s'});
 }
 else {
-    warn "$me: byte, character or field list required\n";
+    warn "byte, character or field list required\n";
     usage();
 }
 exit $rc;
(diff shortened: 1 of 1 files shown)


bin/cut
```diff
@@ -20,8 +20,17 @@ use constant EX_SUCCESS => 0;
 use constant EX_FAILURE => 1;
 
 my $me = basename($0);
-
+my $rc = EX_SUCCESS;
 my %opt;
+
+$SIG{'__WARN__'} = sub {
```
briandfoy, line 26: I don't think I would have done it the same way, but this is fine. I would have probably created a new sub and called that explicitly instead catching every warn. 

But this is fine.

briandfoy (approved):

briandfoy:
changes: ensure the an error condition is set and reported in the exit code

PR #1059, opened 2026-09-17
cmp: same file comparison fix

mknos:
If the two file arguments are the same but the skip values (i.e. offsets) don't match, perform a regular comparison. The shortcut exit case can still be used if the offsets are equal.

Test:
```
%perl cmp cmp cmp 1 2 
cmp cmp differ: char 1, line 1
%
```


briandfoy (approved):

briandfoy:
changes: fix problem comparing a file to itself

